PR #42, opened 2016-08-13
Splitted 'auto' FocusMode into 'single-shot' and 'continuous'

miguelao:
To better reflect what the typical hardware provides.

### DIFF
--- a/index.html
+++ b/index.html
@@ -459,12 +459,14 @@
 
     <section id="FocusMode">
     <h2><code>FocusMode</code></h2>
+    <p>Note that <code>FocusMode</code> is used for both capabilities/status enumeration and for setting options for a capture.</p>
     <div>
       <pre class="idl">
         enum FocusMode {
             "unavailable",
-            "auto",
-            "manual"
+            "manual",
+            "single-shot",
+            "continuous"
         };
       </pre>
     </div>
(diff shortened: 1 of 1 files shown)


miguelao:
Ping, this should be pretty straightforward.

gmandyam:
@miguelao 

If we are going to do this, should we make it compatible with Android HAL - https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#CONTROL_AF_MODE?  There are other modes such as Macro and Extended Depth.

miguelao:
I don't think that's necessary at this moment. Also, note that the focus modes in [Android Camera2 API](https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#CONTROL_AF_MODE) are not exactly one-to-one compatible with the [Android Camera1 API](https://developer.android.com/reference/android/hardware/Camera.Parameters.html#getFocusMode()), nor with the [Mac AVFoundation ones](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVCaptureDevice_Class/#//apple_ref/c/tdef/AVCaptureFocusMode).

We will need to revisit the Focus anyway when we need to add support for metering regions (will make an issue for that).

PR #43, opened 2016-08-16
Fix typo, add example values for focus and flash

dandv:


gmandyam:
Thanks for the corrections.  Will merge.

PR #48, opened 2016-08-22
Remove invalid [Throws] extended attribute

dontcallmedom:
[Throws] is no longer a valid WebIDL extended attribute. In any case, it would not be applicable to a promise-returning function.


gmandyam:
@dontcallmedom 

Thanks for "catch"ing this.

Issue #49, opened 2016-08-23
Revisit ExposureMode and FillLightMode data types to better reflect mobile platforms


miguelao:
Current [ExposureMode](https://w3c.github.io/mediacapture-image/#ExposureMode)s are typical for DSLR cameras (see e.g. [this article](http://www.digital-photo-secrets.com/tip/2879/which-is-best-spot-center-weight-or-matrix-metering/)):
``` 
enum ExposureMode {
    "frame-average",
    "center-weighted",
    "spot-metering"
};
```
whereas [Fill Light Mode](https://w3c.github.io/mediacapture-image/#FillLightMode)s are:
```
enum FillLightMode {
    "unavailable",
    "auto",
    "off",
    "flash",
    "on"
};
```

These don't correlate strongly for mobile platforms, e.g:

- Android Camera2's [AE modes](https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#CONTROL_AE_MODE) and  [flash modes](https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#FLASH_MODE):

``` 
CONTROL_AE_MODE_OFF
CONTROL_AE_MODE_ON
CONTROL_AE_MODE_ON_AUTO_FLASH
CONTROL_AE_MODE_ON_ALWAYS_FLASH
CONTROL_AE_MODE_ON_AUTO_FLASH_REDEYE
```
```
FLASH_MODE_OFF
FLASH_MODE_SINGLE
FLASH_MODE_TORCH
```

- Mac AVFoundation's [exposure mode](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVCaptureDevice_Class/#//apple_ref/c/tdef/AVCaptureDevicePosition). [flash mode] (https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVCaptureDevice_Class/#//apple_ref/c/tdef/AVCaptureFlashMode) and [torch mode](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVCaptureDevice_Class/#//apple_ref/c/tdef/AVCaptureTorchMode):

``` 
typedef enum : NSInteger {
   AVCaptureExposureModeLocked                    = 0,
   AVCaptureExposureModeAutoExpose                = 1,
   AVCaptureExposureModeContinuousAutoExposure    = 2,
   AVCaptureExposureModeCustom                    = 3,
} AVCaptureExposureMode;
```
```
[...]

PR #51, opened 2016-08-23
Generalize FocusMode to MeteringMode and use it for `autoExposureMode`

miguelao:
Proposal to remove the low-level `ExposureMode` and instead use the typical OS level modes, see #49.

Also added `sharpness` entry to `PhotoCapabilities` attributes description, was only in the idl.


miguelao:
Ping.

PR #54, opened 2016-08-31
Converted whiteBalanceMode to MeteringMode and adding colorTemperature

miguelao:
This PR suggest a correction to the (Auto) White Balance getter/setter. `PhotoSettings` says:
```
boolean       autoWhiteBalanceMode;
unsigned long whiteBalanceMode;
```
instead, is better to have `whiteBalanceMode` be a `MeteringMode` (i.e. `manual`, `continuous` etc.) and add a parameter `colorTemperature` to indicate its current/desired magnitude. This `colorTemperature` only applies when `whiteBalanceMode` is in manual mode.

### DIFF
--- a/index.html
+++ b/index.html
@@ -205,13 +205,14 @@
     <div>
       <pre class="idl">
         interface PhotoCapabilities {
-          readonly attribute boolean            autoWhiteBalanceMode;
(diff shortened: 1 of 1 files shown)


miguelao:
Ping.

PR #57, opened 2016-09-07
Use Promise-based MediaDevices.gUM() + cosmetic fixes

### LINKED ISSUE #56: Use consistent MediaDevices vs. navigator .getUserMedia, and the Promise-based API
The three [examples](https://www.w3.org/TR/image-capture/#examples) use a mix of the Promise-based gUM call, and use the [deprecated navigator.getUserMedia](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia).

Should they consistently use `MediaDevices.getUserMedia(...).then(...).catch(...)` ?

dandv:
...also made the whitespace indent consistent (4 spaces). Fix #56.

### DIFF
--- a/index.html
+++ b/index.html
@@ -499,10 +499,10 @@
 
     ##### Grabbing a Frame for Post-Processing
     <pre class='example'>
-    navigator.mediaDevices.getUserMedia({video: true}).then(gotMedia, failedToGetMedia);
+    MediaDevices.getUserMedia({video: true}).then(gotMedia, failedToGetMedia);
 
     function gotMedia(mediastream) {
-        //Extract video track.
+        // Extract video track.
         var videoTrack = mediastream.getVideoTracks()[0];
         // Check if this device supports a picture mode...
         var captureDevice = new ImageCapture(videoTrack);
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -499,10 +499,10 @@
 
     ##### Grabbing a Frame for Post-Processing
     <pre class='example'>
-    navigator.mediaDevices.getUserMedia({video: true}).then(gotMedia, failedToGetMedia);
+    MediaDevices.getUserMedia({video: true}).then(gotMedia, failedToGetMedia);
```
alvestrand: Apologies for missing this, but this is wrong. MediaDevices is the class variable - you can't call MediaDevices.getUserMedia since it's not a static function.

You have to call navigator.mediaDevcies.getUserMedia(), or it won't run.
(If I'm wrong, and you've actually seen code like this running, please provide the example!)

dandv: Np, I was [confused](https://github.com/webrtc/samples/issues/825#issuecomment-245210164) :)

I had also sent another PR, #58.

Issue #64, opened 2016-09-17
Document fails W3C validation


dandv:
A bit ironic, but... https://validator.w3.org/nu/?doc=https%3A%2F%2Fw3c.github.io%2Fmediacapture-image%2Findex.html

gmandyam:
Partially addressed by https://github.com/w3c/mediacapture-image/pull/65/commits/869a301215b28a307832cef128db0fb9f9ae52e1.  Will keep open for now.

gmandyam:
Had to revert https://github.com/w3c/mediacapture-image/pull/65/commits/869a301215b28a307832cef128db0fb9f9ae52e1.  Declarative text does not display properly.

dandv:
Happy to fix as soon as I understand what "declarative text" means.

miguelao:
Only two Warnings remaining:
> Warning: Section lacks heading. Consider using h2-h6 elements to add identifying headings to all sections.

For both:
> `<section id='sotd'>`

and
> `<section id='abstract'>`

Which I think we can live with, so closing this issue perfunctorily.

PR #65, opened 2016-09-17
Fix Example 3, repeated frame grabbing

dandv:
Tested to work in Chrome 55. Also addresses most of #64.


gmandyam:
@dandv  - Thanks for addressing this.

PR #66, opened 2016-09-21
Revert "Fix Example 3, repeated frame grabbing"

gmandyam:
Reverts w3c/mediacapture-image#65.  Example 3 declarative syntax does not display properly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -550,45 +550,42 @@
 
     ##### Repeated grabbing of a frame
     <pre class='example'>
-    &lt;html&rt;
-    &lt;body&rt;
-    &lt;p&rt;&lt;canvas id="frame"&rt;&lt;/canvas&rt;&lt;/p&rt;
-    &lt;button onclick="stopFunction()"&rt;Stop frame grab&lt;/button&rt;
-    &lt;script&rt;
-        var interval;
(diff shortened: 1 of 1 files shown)


gmandyam:
@dandv Please correct and resubmit PR.

dandv:
@gmandyam, can you please clarify on

> Example 3 declarative syntax does not display properly.

My PR executes fine in Chrome 55, whereas the previous example didn't run at all.

Also, if you have a minute, could you add some information to your GitHub profile, since you're a W3C group member? Thanks.

dandv:
@miguelao, any ideas?

gmandyam:
@dandv I have no doubt the example executed properly in Chrome.  The problem was that the text you submitted in the PR was not rendering properly due to what I would imagine was some Respec issue.  When I said "declarative syntax" I meant that the HTML markup (https://en.wikipedia.org/wiki/Declarative_programming) parts of the example were not displaying properly in the final document text.

dandv:
> The problem was that the text you submitted in the PR was not rendering properly

Would you be kind enough to provide a screenshot, or instruction on how to obtain the same display? (The latter could go in the README.)

PR #67, opened 2016-09-26
Fix typo "contructor"

dandv:


miguelao:
@dandv somehow I don't still have superpowers to merge this change, otherwise I'd do it right away since it's trivial. @dontcallmedom can I be invested with such merge-superpowers plz? Thanks

PR #68, opened 2016-09-26
Fix links to ImageCaptureError

dandv:


beaufortfrancois:
@dandv Is that still a thing?

dandv:
@beaufortfrancois: not after #93.

PR #70, opened 2016-09-27
|colorTemperature| should be a MediaSettingsRange

miguelao:
Just noticed during implementation of |colorTemperature| that this field
should be a MediaSettingsRange when considered a capability.


miguelao:
This PR is also trivial so I'm giving it the green light.
(@gmandyam please verify).

PR #81, opened 2016-10-04
Some minor corrections from @fluffy's comments

miguelao:
See individual commits and tagged issues by @fluffy

### DIFF
--- a/index.html
+++ b/index.html
@@ -225,13 +225,13 @@
         <dd>This reflects the current focus mode setting.</dd>
 
         <dt><dfn><code>brightness</code></dfn> of type <span class="idlAttrType"><a>MediaSettingsRange</a></span></dt>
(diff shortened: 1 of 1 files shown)


miguelao:
@fluffy PTAL

fluffy:
I like the changes. I still think that things like brightness need more definition. If I set it to 2, what happens? If I was building a UI that had a slider, what would the range of the slider be ? One possible definition would be it was a value of 2 would roughly double the pixel value given the same light level.

miguelao:
@fluffy added more literature, and a Note detailing that many of these fields cannot be further defined due to them being details of the algorithm which are often hardware secrets.  If you check the UVC standard (USB Device Class Definition for Video Devices) you'll see that the definition of e.g. Brightness control are just as vague:
> 4.2.2.3.2 Brightness Control
This is used to specify the brightness. This is a relative value where increasing values indicate increasing brightness. The MIN and MAX values are sufficient to imply the resolution,

PR #82, opened 2016-10-04
Fix Example 3, repeated frame grabbing

dandv:
Hopefully the document renders correctly now.

Obsoletes #66.


miguelao:
Looks good.  I had to patch it in my repo, and that was hard because there is no way to download a pull request as a diff-patch -- I guess I'm spoiled by using wonderful Rietveld :smile:

PR #95, opened 2016-10-11
Fix copy/pasted grabFrame reference to takePhoto()

dandv:


miguelao:
Argh! Thanks @dandv :+1:

Issue #94, opened 2016-10-11
Rejecting promises vs throwing


dandv:
After [the change from ImageCaptureError to DOMException](https://github.com/w3c/mediacapture-image/pull/93) (which has "also rewritten the 4 method's descriptions to be more Promise-based and homogeneous"), I'd have to update [the polyfill](https://github.com/dandv/imagecapture) to throw a new `DOMException` if the stream is not `live`, but reject in case of other operational errors.

Any reason for this inconsistency which goes again [common recommendations](http://exploringjs.com/es6/ch_promises.html#_tips-for-error-handling)?

> For operational errors, each function should support exactly one way of signaling errors. For Promise-based functions that means not mixing rejections and exceptions, which is the same as saying that they shouldn’t throw exceptions.

miguelao:
Hmmm I guess from the user's POV is best to just have a single handler for the Exceptions coming out of a given method. I was blinded by returning errors as soon as possible :smile: -- would you say is best to always reject-with-DOMException, @dandv ?

dandv:
@miguelao: I would support that, yes.

miguelao:
Done!

PR #96, opened 2016-10-11
Remove DOMException-throws in favour or reject-with-DOMException (#94)

miguelao:


miguelao:
Merging since it's a trivial cleanup.

Issue #107, opened 2016-10-13
Create implementation status page


beaufortfrancois:
It would be great to have a living document that shows the implementation status of the API on the different browsers / platforms like the Web Bluetooth API at https://github.com/WebBluetoothCG/web-bluetooth/blob/gh-pages/implementation-status.md#chrome.

Something like:

Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows |
------------------------- | :-------: | :-----: | :-: | :---: | :-----: |
getPhotoCapabilities()         |  ✓         | ✓        |     |       |         |
└ zoom         |  ✓         | ✓        |     |       |         |
└ brightness         |  55         | ✓        |     |       |         |
takePhoto()         |  ✓         | ✓        |     |       |         |
...

miguelao:
Oh, nice! +1, I'll knock it off.

miguelao:
@beaufortfrancois PTAL and let me know, preventively closing this Issue.

beaufortfrancois:
See nits at https://github.com/w3c/mediacapture-image/pull/110

PR #110, opened 2016-10-14
Polished Implementation status page

beaufortfrancois:
Here are some nits addressed in this patch:
- ordered alphabetically PhotoCapabilities for easy lookup,
- added `pointsOfInterest` row
- moved `grabFrame` to the top
- moved `setOptions` to a separate row
- added `MediaSettingsRange.step` row
FIX: #107


beaufortfrancois:
I've addressed your feedback and add `56` to `MediaSettingsRange.step`. WDYT?

Issue #99, opened 2016-10-12
All examples are invalid


beaufortfrancois:
### https://w3c.github.io/mediacapture-image/#grabbing-a-frame-for-post-processing

- extra parenthesis
- `if (captureDevice) {` is not necessary
- `imgData` is not defined
- `processFrame` can't use `imgData.data` as there as no `data` in `imageBitmap`.

### https://w3c.github.io/mediacapture-image/#taking-a-picture-with-red-eye-reduction-supported-and-used 

- `if (captureDevice) {` is not necessary
- `captureDevice.photoCapabilities.redEyeReduction` is invalid
- `redEyeReductionSetting` should be `redEyeReduction`
- `.then(captureDevice.takePhoto()` is missing a parenthesis
- `img.src = URL.createObjectURL(e.data);` should be `img.src = URL.createObjectURL(blob);`

### https://w3c.github.io/mediacapture-image/#repeated-grabbing-of-a-frame

- `var frameVar;` should be defined outside of this function and maybe even renamed 
- `if (captureDevice) {` is not necessary
- `then(processFrame())` should be `then(processFrame)`
- `function processFrame(e) {` should be `function processFrame(imgData) {`

dandv:
#1 is a known issue - see #62.

miguelao:
1. Doesn't make any sense, since `data` is not existing in `ImageBitmap` - we could just circulate the data through a `<canvas>` to get the same data, wdyt?
2. will be phased out by #103 (addressed in PR #111).
3. Addressed by #102

miguelao:
#114 should have addressed these concerns, ptal @beaufortfrancois .

beaufortfrancois:
It looks like some examples have not been tested at all. Here's a PR that should address this issue: https://github.com/w3c/mediacapture-image/pull/115

miguelao:
I tried to synchronise all the examples back and forth between codepen.io
and the Spec, and #115 is only cosmetic changes IIUC (more consistent
variable naming, etc), but I might have missed something :smile:. 
What's the issue(s) with the current examples?

beaufortfrancois:
You're right about cosmetic changes (consistent variable, `const` instead of `var`, `track` instead of `videoTrack` but there are more:

##### Update camera zoom and <code>takePhoto()</code>

- [ ] `document.querySelector('input[type="img"]');` is invalid. I've replaced it with `document.querySelector('img');`
[...]

PR #115, opened 2016-10-19
Fix examples nits

beaufortfrancois:
FIX: #99

### DIFF
--- a/index.html
+++ b/index.html
@@ -484,8 +484,8 @@
       var imageCapture;
 
       navigator.mediaDevices.getUserMedia({video: true})
-          .then(gotMedia)
-          .catch(e => { console.error('getUserMedia() failed: ', e); });
+        .then(gotMedia)
+        .catch(err => console.error('getUserMedia() failed: ', err));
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -506,50 +509,51 @@
...
-          .then((blob) => {
-            console.log("Photo taken: " + blob.type + ", " + blob.size + "B")
-
-            const imageTag = document.querySelector('input[type="img"]');
-            imageTag.src = URL.createObjectURL(blob);
-          })
-          .catch((err) => {
-            console.error("takePhoto() failed: ", e);
-          });
+        imageCapture.takePhoto().then(blob => {
```
miguelao: I'm trying to put the .then() in a new line,
indented by four spaces, like in l.488, 599 etc. 
Let's pick one and be consistent?

beaufortfrancois: You're right. Sorry for that.

beaufortfrancois:
@miguelao I've updated the PR. WDYT?

miguelao:
I sayz good! Merging it.

PR #116, opened 2016-10-19
MediaSettingsRange.step shows up in M56

beaufortfrancois:
If I'm not mistaken, `MediaSettingsRange.step` was introduced in https://chromium.googlesource.com/chromium/src.git/+/8f17f7e85f0c37a5d6549ea252baa15b9653a974
According to https://chromium.googlesource.com/chromium/src/+/8f17f7e85f0c37a5d6549ea252baa15b9653a974/chrome/VERSION, Chrome version is:
```
MAJOR=56
MINOR=0
BUILD=2891
PATCH=0


miguelao:
You're right, concretely in 56.0.2891.8, I got mislead by the Dev channel in my Chromebook.

Issue #97, opened 2016-10-12
Make PhotoCapabilities object keys enumerable


beaufortfrancois:
It would be nice if we could simply use `JSON.stringify` on the object returned by `getPhotoCapabilities`.

```js
imageCapture.getPhotoCapabilities().then(photoCapabilities => {
  console.log(JSON.stringify(photoCapabilities));
});
```

For now, it simply returns `{}`

beaufortfrancois:
gentle ping @miguelao

miguelao:
I agree with this idea, I wonder: Should this be done by means of an [`iterable`](https://heycam.github.io/webidl/#dfn-iterable-declaration) declaration in the `interface PhotoCapabilities`? Took a look around the Chromium code base and there are few examples of `<DOMString>` though.

beaufortfrancois:
@jyasskin might help there.

jyasskin:
Check out https://heycam.github.io/webidl/#idl-serializers for the way to define the `toJSON()` method.

miguelao:
Thanks @jyasskin! Will work on a PR verys shortly, in the meantime https://crrev.com/2444433002 has the Chrome patch.

miguelao:
@beaufortfrancois, during review of the [Chrome CL](https://codereview.chromium.org/2444433002/#msg7) @domenic raised an interesting concern about this extension, let me reproduce it here:
> In general serializers are used for data structures that the spec author
envisions being commonly transported and serialized via JSON. We don't add them
to anything that could conceivably be serialized, in general; we add them for
things that we plausibly expect authors to send over the wire.

Do we want the extra bloat, or are we OK with writing a JS snippet to generate a string from a  `PhotoCapabilities`?

beaufortfrancois:
Concern is definitely valid.

For info, I found myself needing the `toJSON` binding when exploring the API and my intent was to show raw data from the API.

domenic:
In general that's not a good reason to add toJSON methods. It's rather a reason to improve your browser's `console.log` to do so (as it does in e.g. Chrome).

miguelao:
@beaufortfrancois , in light of the comments, do we still need a `serializer` ? Otherwise we could close this bug.

beaufortfrancois:
Let's close this.

PR #118, opened 2016-10-21
Add 'serializer = {attribute};' to PhotoCapabilities and MediaSettingsRange (#97)

miguelao:
This PR addresses #97, if we still deem it appropriate.


miguelao:
#97 is closed, and so is this PR.

Issue #77, opened 2016-10-04
types of setting objects


fluffy:
Attributes such as exposureCompensation would be better as a double than a long x100. I can not see any disadvantage to using a double here.

gmandyam:
@miguelao 

Any objection to the suggestion of @fluffy?

miguelao:
@gmandyam, if you're ok I'm ok.
I'd suggest making it `float` which is generally as large as `int` (vs `double` which is usually larger) and I don't see any need for the extra ultra-precision.
I can write the PR then.

miguelao:
#120 is the PR, @fluffy, @gmandyam, PTAL

dontcallmedom:
@miguelao [`float` is discouraged](https://heycam.github.io/webidl/#idl-float):
> Unless there are specific reasons to use a 32 bit floating point type, specifications should use double rather than float, since the set of values that a double can represent more closely matches an ECMAScript Number.

miguelao:
@dontcallmedom, that's exactly what I mentioned in https://github.com/w3c/mediacapture-image/pull/120#issue-187468145, but still "I feel that float can represent all of our value ranges and is smaller in size." Which one will win?

dontcallmedom:
but it's not really smaller in size - the underlying ecmascript engine only sees double.

miguelao:
Closed per #120.

miguelao:
Landed in chrome, see https://crbug.com/663021.

PR #120, opened 2016-11-05
(#77) s/long/double/ in MediaSettingsRange

miguelao:
...and removed entries mentioning a x100 multiplier to avoid floating point.

Note that the [float](https://heycam.github.io/webidl/#idl-float) idl definition mentions:
> Unless there are specific reasons to use a 32 bit floating point type, specifications should use double rather than float, since the set of values that a double can represent more closely matches an ECMAScript Number.

but I feel that `float` can represent all of our value ranges and is smaller in size.


miguelao:
s/float/double/ per comments in #77

Issue #124, opened 2016-12-07
Use the constrainable pattern instead of creating a subtly different function set


alvestrand:
The combination of getPhotoCapabilities (with both ranges and current settings) and setOptions creates an almost equivalent set of functionalities to the constrainable pattern used on MediaStreamTrack.

Two nearly identical patterns is a bad thing for the platform. Please reconsider.

(This has been discussed before, and I know giri does not agree with me. I wish for the discussion to be recorded as a top-level issue in this tracker.)

stefhak:
Asking in particular (I think) @gmandyam and @miguelao : what would the drawback(s) be of aligning with mediacapture-main and using the constrainable pattern?

gmandyam:
@stefhak 

I did provide my view on applicability of constraints to ImageCapture in the discussion on https://github.com/w3c/mediacapture-image/issues/73.  In addition, I would note the following.  Apologies for any clumsy wording.

a) The contraints mechanism is already leveraged in ImageCapture, as the constructor (https://w3c.github.io/mediacapture-image/#ImageCaptureAPI)  uses a track to which constraints have been applied. 

b) That being said, constraints (in my opinion) add value when there is some form of contention.  By 'contention', I am describing a situation where for instance multiple tabs in the active browser context require media capture, each with their own individual requirements.  However, for ImageCapture this kind of contention is not applicable as the ImageCapture object is constructed using a track that has already been created based on the browser constraint resolution.

c) It is still unclear (to me at least) whether developers have fully understood the constraints concept in the context of MediaStreams and actually found value in using them.  In that respect, I am concerned about causing developer confusion by adding a constraints-on-top-of-constraints mechanism.  For instance, developers may ask why they cannot apply image capture constraints directly to the media stream itself.

stefhak:
@gmandyam I had not seen #73, thanks for the pointer.

A related question: has the core part of #73 (i.e. that some of the capabilities reported via ```getPhotoCapabilities()``` are actually settings) been addressed? Or is there no need (e.g. it is OK to report a setting because the capabilities are evident and there is never a need to probe for them)?

jan-ivar:
> b) That being said, constraints (in my opinion) add value when there is some form of contention. 

@gmandyam If two tabs are sharing the same camera and one calls:
```js
new ImageCapture(stream.getVideoTracks()[0]).setOptions({ zoom: 4 });
```
and the other calls:
```js
[...]

PR #146, opened 2017-01-21
Use constrainable pattern for ImageCapture (#124)

miguelao:
First draft of using [`ConstrainablePattern`](https://www.w3.org/TR/mediacapture-streams/#constrainable-interface) for ImageCapture capabilities and settings management.

This PR:
- Adds an `Extensions` section, where `MediaTrackSupportedConstraints`, `MediaTrackCapabilities`, `MediaTrackConstraintSet` and `MediaTrackSettings` dictionaries are extended to include our set of capabilities/settings. 
- The previous partial dictionaries are also explained in relation to a new 'Constrainable properties' section (since these dictionaries always refer to the same set of constraints/settings, it's easier to define them all together).
- moves `getPhotoCapabilities()`, `setOptions()`, `PhotoCapabilities` and `PhotoSettings` to a legacy non-normative section, for the time being.

A rendering of the contents of this PR can be found in https://rawgit.com/Miguelao/mediacapture-image/pr001_constrainable_pattern/index.html


miguelao:
@alvestrand can you please take a first coarse look?

beaufortfrancois:
@miguelao Can you update examples as well?

index.bs
```diff
@@ -40,7 +40,7 @@ table td, table th {
 
 The API defined in this document captures images from a photographic device referenced through a valid {{MediaStreamTrack}}. The produced image can be in the form of a {{Blob}} (see {{takePhoto()}} method) or as a {{ImageBitmap}} (see {{grabFrame()}}). Picture-specific settings can be optionally provided as arguments that can be applied to the device for the capture.
 
-# Image Capture API # {#imagecaptureapi"}
+# Image Capture API # {#imagecaptureapi}
```
alvestrand: You need an explanaition up here to explain where the constraints fit in. Something like:

"This document defines a number of new constraints for a MediaStreamTrack that can be applied in order to make its behavior more suitable for taking pictures. The use of these constraints is via the standard getUserMedia() and applyConstraints() methods for a MediaStreamTrack. They will modify the behavior of the track and its source independently of the usage of the track with the ImageCapture object."

index.bs
```diff
@@ -114,224 +112,272 @@ interface ImageCapture {
...
-  The supported resolutions are presented as segregated {{imageWidth}} and {{imageHeight}} ranges to prevent increasing the fingerprinting surface and to allow the UA to make a best-effort decision with regards to actual hardware configuration.
+{{MediaTrackSettings}} can be retrieved to verify the effect of the application by the user agent of the requested {{MediaTrackConstraints}}. Some constraints such as, e.g. <a>zoom</a>, might not be immediately applicable.
 </div>
 
-## Discussion
-<em>This section is non-normative.</em>
+<pre class="idl">
+  partial dictionary MediaTrackConstraintSet {
+    MeteringMode       whiteBalanceMode;
+    ConstraintDouble   colorTemperature;
```
miguelao: Corrected. The link was working because I defined it at the bottom of the file, and the `text` section was wrong while the `url` was right:
> `urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: interface; text: ConstrainDouble ; url: idl-def-ConstrainDouble
`

index.bs
```diff
@@ -114,224 +112,272 @@ interface ImageCapture {
...
-  </table>
+### Members ### {#mediatracksupportedconstraints-members}
 
-  <div class="note">
-    Many of the {{PhotoSettings}} represent hardware capabilities that cannot be modified instaneously, e.g. `zoom` or `focus`. {{setOptions()}} will resolve the Promise as soon as possible. The actual status of any field can be monitored using {{getPhotoCapabilities()}}.
-  </div>
-  </dd>
+<dl class="domintro">
+  <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>whiteBalanceMode</code></dfn></dt>
+  <dd>Whether <a>white balance mode</a> constraining is supported.</dd>
```
miguelao: Done.

index.bs
miguelao: Done.

index.bs
```diff
@@ -352,75 +398,126 @@ interface ImageCapture {
...
+
+    <li><dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens.  The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1.  The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).</li>
+
+    <li><dfn>Fill light mode</dfn> describes the flash setting of the capture device (e.g. |auto|, |off|, |on|). </li>
+  </ol>
+
+# {{MediaSettingsRange}} # {#mediasettingsrange-section}
 
 <pre class="idl">
   interface MediaSettingsRange {
```
miguelao: MediaSettingsRange is an interface 
```
<pre class="idl">
  interface MediaSettingsRange {
      readonly attribute double max;
      readonly attribute double min;
      readonly attribute double step;
  };
</pre>
```
whereas [`DoubleRange`](https://www.w3.org/TR/mediacapture-streams/#idl-def-DoubleRange) is a dictionary.
 
```
dictionary DoubleRange {
             double max;
             double min;
};
```
It makes sense to have it as a Dictionary since it's used for configuring constraints in the track; in ImageCapture it makes sense to have it as an interface because settings ranges are always composed of min-max and step.  So, they're different ways to represent related concepts and both are OK.

miguelao:
@beaufortfrancois the examples as they are would still work assuming that `ImageCapture` is also a `LegacyImageCapture`, which still has `getPhotoCapabilities()` and `setOptions()`.  I propose making new examples and making evident that the current one is legacy-oriented in another, subsequent CL (if we could only make GitHub issues "blocked" on others as we do in crbug.com... then it'd be clearer).  Any way I filed #148 for it.

miguelao:
@alvestrand comments addressed except the `MediaSettingsRange` one. PTAL.

@gmandyam I think you can take a look now, PTAL.

miguelao:
ping @alvestrand

miguelao:
@alvestrand @gmandyam PTAL

gmandyam:
@miguelao 
Why are the ConstrainDouble and ConstraintBoolean types necessary?  It is hard to trace exactly what data types the corresponding constraints must follow.

miguelao:
`ConstrainBla` allows for things like e.g. `{ zoom : 215 }`, { zoom : { min : 200, max : 225}}`, `{ zoom : { exact : 213 }}` etc which  are part of the constrainable pattern definition. (Note that there's a typo, `ConstraintBoolean` should be `ConstrainBoolean`).

gmandyam:
Sorry - that was my mistake.  I thought you had redefined ConstrainBoolean, etc. in the Image Capture spec.  Turns out you had just linked to the Media Capture and Streams spec.

gmandyam:
Assume that the MediaStream is currently sourcing a <video> element.  Upon a successful call of applyConstraints(), is the expectation that all the media track constraints defined in Image Capture will be reflected in the video being rendered in the <video> element (e.g. contrast, sharpness, etc.)?  Or will they only be applied when one of the ImageCapture methods (e.g. takePhoto) are invoked?

miguelao:
I believe they will be applied and visible at the application moment (plus any hardware delays, e.g. zoom) since we'll be modifying the Track constraints.

miguelao:
@alvestrand ping
@gmandyam all clear now from your side?

gmandyam:
@miguelao, @alvestrand 

I think we can go forward with this change.

miguelao:
@alvestrand please take another look otherwise it's good to go.

index.bs
```diff
@@ -352,75 +400,126 @@ interface ImageCapture {
...
-  <dd>This reflects the desired fill light (flash) mode setting.  Acceptable values are of type {{FillLightMode}}.</dd>
+  <dt><dfn dict-member for="MediaTrackSettings"><code>fillLightMode</code></dfn></dt>
+  <dd>This reflects the current <a>fill light mode</a> setting (flash).</dd>
 
 </dl>
 
-# MediaSettingsRange
+## Constrainable Properties ## {#constrainable-properties}
+
+  Many of these fields mirror hardware capabilities that are hard to define since can be implemented in a number of ways.  Moreover, hardware manufacturers tend to publish vague definitions to protect their intellectual property.  The following definitions are assumed for individual settings and are provided for information purposes:
```
stefhak: nit: should there be a ```they``` or something similar inserted between ```since``` and ```can```?

stefhak:
@miguelao Harald is on vacation for the next two weeks. Unless @burnburn can review and comment quickly I propose you merge and that we take it from there. @burnburn would you have the cycles to go over this in the next few days?

stefhak:
It seems @burnburn does not have the time to review now, I think this PR should be merged now.

jan-ivar:
I thought we were [only going to move](https://github.com/w3c/mediacapture-image/issues/124#issuecomment-268377511) "[live](https://github.com/w3c/mediacapture-image/issues/124#issuecomment-268318169)" settings to constraints? This appears to move all of them.

jan-ivar:
That may be fine, just checking if I missed the rationale.

miguelao:
@jan-ivar I thought about it, but it might be too complex to have two ways to configure the device; the quintessential example are `imageHeight` and `imageWidth`: they are not (visually) reflected in the Track, but seems logical that the user would expect them listed when calling `getSettings()` or `getCapabilities()`, etc.  IOW,  I think it might be better to have them all together than spread across two interfaces.  wdyt?

jan-ivar:
@miguelao Users would be wrong. :-) What we call "delayed" properties aren't really applied to the device at all at the time of `applyConstraints`, but later. Thus `getSettings()` would  reveal no current hardware settings, rather just mirror `getConstraints()`. Doesn't seem very useful.

Also, there'd never be any conflict; no need for `min`, `max` or `exact` except for purposely going outside the deterministic `getCapabilities()` envelope.

 The alternative seems more logical to me:

    capture.takePhoto({imageWidth: 3264, imageHeight: 2448});

I.e. fess up to these not really being video track properties, but properties of the `takePhoto` method.

Of course, logic isn't everything. Expecting users to remember which settings are video settings and which are photo settings, may frustrate more than the functional clarity helps.

jan-ivar:
But if `imageWidth`, `imageHeight` were `takePhoto` options, they wouldn't belong in `getCapabilities`.

miguelao:
@jan-ivar I understand your last 2 comments; they'd apply to `imageWidth`, `imageHeight` and `redEyeReduction` since those are the only ones not "visible" / "live" (only applied when the photo is actually taken and not reflected on the MSVideoTrack).

But since those 3 still have ranges/supported sets (e.g. `imageWidth` might support any value between 400px and 4000px in steps of 100px, `redEyeReduction` can support `{true, false}`, how would we convey those ranges to the WebApp? Via `getCapabilities()`? Via some reduced `getPhotoCapabilities()`? (And a similar argument goes for configuring them). Just double checking that I understood the concrete proposal.

miguelao:
ping @jan-ivar

jan-ivar:
@miguelao Isn't there also `exposureMode`, `iso`, `focusMode`, and `fillLightMode` from your [list](https://github.com/w3c/mediacapture-image/issues/124#issuecomment-268318169)?

I don't have a concrete proposal, merely challenging what is most logical:
 * Separate video and photo settings is more logical, less convenient.
 * Single mechanism for settings is more convenient, less logical.

If we chose the former, any ranges for values of `takePhoto` arguments would not belong in `getCapabilities`, and would need their own thing IMHO, e.g. `getPhotoCapabilities()`.

>e.g. imageWidth might support any value between 400px and 4000px in steps of 100px

In a vacuum where constraints didn't exist, I'd say the most logical way to represent "delayed" (photo) settings would be an array of values for each thing, and then have `takePhoto` fail if called with values not in the array. No need for a separate "configuring" step, since you know all the valid settings. You're not configuring the device, you're configuring the shot.

If we chose the latter, for "delayed" (photo) properties which never collide, the `exact` keyword seems to exist solely to compensate for the lack of specificity in `getCapabilities()`.

I'm curious, how do we imagine a user "configuring" their device in the single constraints model? What would the UI look like? How would the app learn the discrete settings to offer as choices?

* Call `applyConstraints({imageWidth: {exact: w}, imageHeight: {exact: h}})` until it succeeds?
* Call `applyConstraints({imageWidth: {min: minWidthICanAccept, max: maxWidthICanAccept}, imageHeight: {min: minHeightICanAccept, max: maxHeightICanAccept}})` and hope it doesn't fail?

What's the ideal UI here? Maybe we should start there.

miguelao:
Let's bring the list here for conveniency (apologies, `iso` can be seen immediately in the video feed).

| Setting | |
|---|---|
| whiteBalanceMode | immediate |    
| colorTemperature | immediate |
| exposureMode | delayed |
| exposureCompensation |  immediate |
| iso | **immediate** | 
| redEyeReduction | delayed |
| focusMode | delayed |
| pointsOfInterest | probably immediate |
| brightness | immediate |
| contrast | immediate |
| saturation | immediate |
| sharpness |  immediate |
| imageHeight | delayed |
| imageWidth | delayed
| zoom | immediate |
| fillLightMode | both |        

Code wise we could 
1. keep the current `Promise<PhotoCapabilities> getPhotoCapabilities();`   returning the "delayed" capabilities (might need some tweaking), i.e. 
```
interface PhotoCapabilities {
  readonly attribute MeteringMode       exposureMode;
  readonly attribute boolean            redEyeReduction;
  readonly attribute MeteringMode       focusMode;
  readonly attribute MediaSettingsRange imageHeight;
  readonly attribute MediaSettingsRange imageWidth;
  readonly attribute FillLightMode      fillLightMode;
};
```
2. nuke the current `setOptions()` and instead pass the mentioned Capabilities directly into `takePhoto()`, i.e. `takePhoto({redEyeReduction : yes, fillLightMode : "auto"});`

3. All the other settings/capabilities would be inspected and/or set via the MediaStreamTrack, e.g. ( `iso` only for clarity):
`track.getCapabilities()` ==> `{ iso : { min:1000, max : 8000, step : 100, default : 3000}}` 
then
`track.applyConstraints({ iso : { ideal:1234}});` ==> works or not
`track.getSettings()` ==> `{iso : 1234}`
 

So essentially this is the `Separate video and photo settings is more logical, less convenient.` case of yours.  If you agree, I will make a separate PR with this approach extended so we can decide on actualities, wdyt?

alvestrand:
This makes sense to me - from the viewpoint of "are these constraints usable for video streams", they all seem usable. Agree that "pointsOfInterest" is likely to be immediate & useful for video.

WRT the odd "fillLightMode" case (both), would it make sense to separate this into two capabilities, "light" (a constraint) and "flash" (a photo property)?

miguelao:
Two updates: I just noticed that `exposureMode` and`focusMode` are also 'live' capabilities, since from the moment the user configures them to any of their values (e.g. `off`, `single-shot`, `continuous`) the effect is immediately visible (as is e.g. `whiteBalanceMode`). Apologies for any confusion.  

I like @alvestrand suggestion, so I'll split `torch` mode out of `fillLightMode`, the latter being a photo-capability and the former being a video constraint.

So the only remaining photo capabilities will be:

```
interface PhotoCapabilities {
  readonly attribute FillLightMode      fillLightMode;  //not including 'torch'
  readonly attribute MediaSettingsRange imageHeight;
  readonly attribute MediaSettingsRange imageWidth;
  readonly attribute FillLightMode      fillLightMode;
};
```

Finalising a new pull request so we can continue the conversation there.

miguelao:
#150 is the new pull request.

jan-ivar:
> track.getCapabilities() ==> { iso : { min:1000, max : 8000, step : 100, default :

👍 except note that there's no such thing as `step:` in constrainable capabilities (e.g. [LongRange](https://w3c.github.io/mediacapture-main/getusermedia.html#idl-def-longrange)).

miguelao:
Closing this in favour of #150 .

PR #150, opened 2017-03-03
Use constrainable pattern for ImageCapture (second take)

miguelao:
Second draft of adding-constrainable-pattern; two ways of configuring are considered:
- "live" capabilities, the result of which modification is reflected in the source MediaStreamTrack
- "non-live" or "photo" capabilities, the result of which modification is only visible in any taken picture.

Also I extracted the `torch` mode (on,off, being "live") out of `fillLightMode` (flash on, off, auto) which is "non-live".

Live rendering in https://rawgit.com/Miguelao/mediacapture-image/pr003_constrainable_pattern__alternative/index.html

This PR branches off #146 at the height of https://github.com/w3c/mediacapture-image/pull/146#issuecomment-284028751


index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
-    "unavailable",
-    "auto",
-    "off",
-    "flash",
-    "torch"
+  partial dictionary MediaTrackSettings {
+    MeteringMode      whiteBalanceMode;
+    MeteringMode      exposureMode;
+    MeteringMode      focusMode;
+    sequence&lt;Point2D> pointsOfInterest;
```
jan-ivar: The > needs escaping.

Also, a sequence of Point2D would be an entirely new constraint. Constraining on it would need to be worked out. You'd probably need to define new Constrain* types for it. The closest to model on may be ConstrainDOMString, which is array-based.

I also confess I don't know what pointsOfInterest is for here in practice, or how it applies to a video stream. How is it intended to be used?

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
+    ConstrainDouble    iso;
+
+    ConstrainDouble    brightness;
+    ConstrainDouble    contrast;
+    ConstrainDouble    saturation;
+    ConstrainDouble    sharpness;
+
+    ConstrainDouble    zoom;
+
+    ConstraintBoolean  torch;
```
jan-ivar: s/ConstraintBoolean/ConstrainBoolean/

index.bs
```diff
@@ -38,36 +38,38 @@ table td, table th {
...
 
 The User Agent must support Promises in order to implement the Image Capture API.  Any {{Promise}} object is assumed to have a resolver object, with <code>resolve()</code> and <code>reject()</code> methods associated with it.
 
 <pre class="idl">
 [Constructor(MediaStreamTrack track)]
 interface ImageCapture {
-   readonly attribute MediaStreamTrack videoStreamTrack;
-
-   Promise&lt;Blob&gt;              takePhoto();
+   Promise&lt;Blob&gt;         takePhoto(optional PhotoSettings photoSettings);
```
miguelao: Ugh yeah, sorry, thanks.

index.bs
```diff
@@ -38,36 +38,38 @@ table td, table th {
...
    Promise&lt;PhotoCapabilities&gt; getPhotoCapabilities();
-   Promise&lt;void&gt;              setOptions(optional PhotoSettings photoSettings);
-   Promise&lt;ImageBitmap&gt;       grabFrame();
+
+   Promise&lt;ImageBitmap&gt;  grabFrame();
```
miguelao: Mere sloppiness: corrected.

index.bs
```diff
@@ -38,36 +38,38 @@ table td, table th {
 
 # Introduction # {#introduction}
 
-The API defined in this document captures images from a photographic device referenced through a valid {{MediaStreamTrack}}. The produced image can be in the form of a {{Blob}} (see {{takePhoto()}} method) or as a {{ImageBitmap}} (see {{grabFrame()}}). Picture-specific settings can be optionally provided as arguments that can be applied to the device for the capture.
+The API defined in this document captures images from a photographic device referenced through a valid {{MediaStreamTrack}}. The produced image can be in the form of a {{Blob}} (see {{takePhoto()}} method) or as a {{ImageBitmap}} (see {{grabFrame()}}).
 
-# Image Capture API # {#imagecaptureapi"}
+Reading capabilities and settings and applying constraints is done in one of two ways depending on if it impacts the source {{MediaStreamTrack}} or not.  Picture-specific capabilities, default values and current settings can be retrieved via {{getPhotoCapabilities()}} and configured via {{takePhoto()}}'s {{PhotoSettings}} argument.  Manipulating video-related capabilities, current settings and constraints is done via the MediaStreamTrack <a href="#extensions">extension mechanism</a>.
```
miguelao: > This introduces the distinction between video settings vs. photo settings.

Correct and intended, I wanted to write about this upfront so any potential reader will notice this duality and why it's like that.

>Maybe the sentence would be even clearer if we s/source MediaStreamTrack/video MediaStreamTrack/ and s/Picture/Photo/, and maybe s/if/whether/ ?

Done all three.

index.bs
```diff
@@ -92,15 +94,17 @@ interface ImageCapture {
   </table>
   </dd>
 
-  <dt><dfn method for="ImageCapture"><code>takePhoto()</code></dfn></dt>
+  <dt><dfn method for="ImageCapture"><code>takePhoto(optional PhotoSettings photoSettings)</code></dfn></dt>
   <dd>
-  {{takePhoto()}} produces the result of a single photographic exposure using the video capture device sourcing the {{videoStreamTrack}}, applying any {{PhotoSettings}} previously configured, and returning an encoded image in the form of a {{Blob}} if successful. When this method is invoked:
+  {{takePhoto()}} produces the result of a single photographic exposure using the video capture device sourcing the {{videoStreamTrack}} and including any {{PhotoSettings}} previously configured, returning an encoded image in the form of a {{Blob}} if successful. When this method is invoked:
```
miguelao: Done.

index.bs
```diff
@@ -92,15 +94,17 @@ interface ImageCapture {
...
   <ol>
     <li>If the {{readyState}} of {{videoStreamTrack}} provided in the constructor is not {{live}}, return <a>a promise rejected with</a> a new {{DOMException}} whose name is {{InvalidStateError}}. </li>
 
+    <li>If an invalid {{PhotoSettings}} object is passed as argument, return <a>a promise rejected with</a> a new {{DOMException}} whose name is {{SyntaxError}}</li>
```
miguelao: Makes sense, removed this paragraph then.

index.bs
```diff
@@ -167,20 +130,31 @@ interface ImageCapture {
...
-  </div>
+  </dd>
+
+  <dt><dfn method for="ImageCapture"><code>getPhotoCapabilities()</code></dfn></dt>
+  <dd>
+  When {{getPhotoCapabilities()}} is used to retrieve the ranges of available configuration options and their current setting values, if any. When this method is invoked:
+  <ol>
+    <li>If the {{readyState}} of {{videoStreamTrack}} provided in the constructor is not {{live}}, return <a>a promise rejected with</a> a new {{DOMException}} whose name is {{InvalidStateError}}. </li>
+
+    <li>Otherwise it MUST queue a task, using the DOM manipulation task source, that runs the following steps:
```
miguelao: Yeah, I'm not sure why it isn't there. Added now.

index.bs
```diff
@@ -167,20 +130,31 @@ interface ImageCapture {
...
   </table>
 
-  <div class="note">
-    Many of the {{PhotoSettings}} represent hardware capabilities that cannot be modified instaneously, e.g. `zoom` or `focus`. {{setOptions()}} will resolve the Promise as soon as possible. The actual status of any field can be monitored using {{getPhotoCapabilities()}}.
-  </div>
+  </dd>
+
+  <dt><dfn method for="ImageCapture"><code>getPhotoCapabilities()</code></dfn></dt>
+  <dd>
+  When {{getPhotoCapabilities()}} is used to retrieve the ranges of available configuration options and their current setting values, if any. When this method is invoked:
```
miguelao: Done.

index.bs
```diff
@@ -167,20 +130,31 @@ interface ImageCapture {
...
+  <ol>
+    <li>If the {{readyState}} of {{videoStreamTrack}} provided in the constructor is not {{live}}, return <a>a promise rejected with</a> a new {{DOMException}} whose name is {{InvalidStateError}}. </li>
+
+    <li>Otherwise it MUST queue a task, using the DOM manipulation task source, that runs the following steps:
+    <ol>
+      <li>Gather data from {{videoStreamTrack}} into a {{PhotoCapabilities}} object containing the available capabilities of the device, including ranges where appropriate. The resolved {{PhotoCapabilities}} will also include the current conditions in which the capabilities of the device are found. The method of doing this will depend on the underlying device. </li>
+
+      <li>If the UA is unable to execute the {{getPhotoCapabilities()}} method for any reason (for example, the {{MediaStreamTrack}} being ended asynchronously), then the UA MUST return <a>a promise rejected with</a> a new {{DOMException}} whose name is {{OperationError}}.</li>
+
+      <li>Return a resolved promise with the {{PhotoCapabilities}} object.</li>
```
miguelao: Done here and elsewhere.

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
   };
 </pre>
 
 ## Attributes ## {##photocapabilities-attributes}
 
 <dl class="domintro">
-  <dt><dfn attribute for="PhotoCapabilities"><code>whiteBalanceMode</code></dfn></dt>
-  <dd>This reflects the current <a>white balance mode</a> setting. </dd>
+  <dt><dfn attribute for="PhotoCapabilities"><code>redEyeReduction</code></dfn></dt>
+  <dd>This reflects whether <a>red eye reduction</a> is suported, and is boolean - true means supported.</dd>
```
miguelao: It's unlikely there isn't some switch to disable the red-eye reduction because it implies a pre-flash that takes a significant amount of time compared to the exposure time; however it might be a good insurance policy for the future. In that case it could read, inspired by `echoCancellation`:

```
If the source cannot do <a>red eye reduction</a> a single false is reported. If 
<a>red eye reduction</a> cannot be turned off, a single true is reported. If the 
script can control the feature, the source reports a list with both true and false 
as possible values.
```
with the interface entry being
`readonly attribute sequence<boolean> redEyeReduction;`

other opinions, @gmandyam ?

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
 
-    <div class="note">
-      <a>Brightness</a>, <a>contrast</a>, <a>saturation</a> and <a>sharpness</a> are specified in [[UVC]].
-    </div></li>
+# Extensions # {#extensions}
 
-    <li><i>Zoom</i> is a numeric camera setting that controls the focal length of the lens.  The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1.  The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).</li>
+This Section defines a number of new constraints for {{MediaStreamTrack}} that can be applied in order to make its behavior more suitable for taking pictures.  Use of these constraints via {{MediaStreamTrack}}'s methods {{getCapabilities()}}, {{getSettings()}}, {{getConstraints()}} and {{applyConstraints()}} will modify the behavior of the track and its source independently of the usage of the {{videoStreamTrack}} with the {{ImageCapture}} object.
```
miguelao: Rewritten for more clarity and linked to the constrainable pattern.

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
 
-  <dt><dfn dict-member for="PhotoSettings"><code>fillLightMode</code></dfn></dt>
-  <dd>This reflects the desired fill light (flash) mode setting.  Acceptable values are of type {{FillLightMode}}.</dd>
+<pre class="idl">
+  partial dictionary MediaTrackCapabilities {
+    sequence&lt;MeteringMode>  whiteBalanceMode;
+    sequence&lt;MeteringMode>  exposureMode;
+    sequence&lt;MeteringMode>  focusMode;
+
+    MediaSettingsRange      exposureCompensation;
```
miguelao: Yeah, the `MediaSettingsRange` was introduced in this Spec to mimic the UVC specification [1] where the properties have max, min and step. Also in this line, developers/devrels requested the `default` value, but that's another discussion 😉 


[1] buried under http://www.usb.org/developers/docs/devclass_docs/

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
-  interface MediaSettingsRange {
-      readonly attribute double max;
-      readonly attribute double min;
-      readonly attribute double current;
-      readonly attribute double step;
+  partial dictionary MediaTrackConstraintSet {
+    MeteringMode       whiteBalanceMode;
+    MeteringMode       exposureMode;
+    MeteringMode       focusMode;
+    ConstraintBoolean  pointsOfInterest;
```
miguelao: Done s/ConstraintBoolean/ConstrainBoolean/

This constraint represents support or not for `pointsOfInterest`.

index.bs
miguelao: Done.

index.bs
miguelao: My idea here was to pass a series of `Point2D`s to be interpreted as `exact`, but I'm no constrainable-pattern-guru; any ideas @alvestrand ?

`pointsOfInterest` is used, if present, to indicate area(s) of the image to use as seeds the 3A (auto-focus, auto-exposure, auto-white balance).  If they are not specified (unconstrained), the whole sensor (or a subsampled version) is used instead.

index.bs
```diff
@@ -206,268 +177,455 @@ interface ImageCapture {
...
+</div>
 
 <pre class="idl">
-  interface MediaSettingsRange {
-      readonly attribute double max;
-      readonly attribute double min;
-      readonly attribute double current;
-      readonly attribute double step;
+  partial dictionary MediaTrackConstraintSet {
+    MeteringMode       whiteBalanceMode;
```
miguelao: Hmm is there no easy way to at least link between enum values and constraint? E.g. by specifying that the [`ConstrainDOMString`](https://w3c.github.io/mediacapture-main/getusermedia.html#idl-def-constraindomstring) needs to derive the accepted values from the said `enum`-->`toString()` ? @alvestrand, ideas?

index.bs
jan-ivar: Yeah I lobbied for `ideal` to mean default value in capabilities long ago in mediacapture, but there was no interest at the time. Shouldn't disrupt anything IMHO.

index.bs
jan-ivar: We can link it in prose in the spec like we did for [facingMode](https://w3c.github.io/mediacapture-main/getusermedia.html#def-constraint-facingMode). Enums are just strings in practice anyway.

index.bs
jan-ivar: @miguelao Note that MediaTrackConstraintSet is effectively the dictionary passed to `applyConstraints`, so a boolean here leaves us no way to actually set the Point2D array.

index.bs
miguelao: Done adding it.

index.bs
miguelao: Cool, changed to `ConstrainDOMString` and added an explanatory line `This string MUST be one of the members of {{MeteringMode}}. `

index.bs
```diff
@@ -206,268 +175,455 @@ interface ImageCapture {
...
+    MeteringMode       exposureMode;
+    MeteringMode       focusMode;
+    ConstrainBoolean   pointsOfInterest;
```
miguelao: @jan-ivar said:
> @miguelao Note that MediaTrackConstraintSet is effectively the dictionary passed to applyConstraints, so a boolean here leaves us no way to actually set the Point2D array.

(I got it via e-mail, but the comment seems lost in this interface :?, hence I copied it here again).

Changed to `sequence<Point2D>`;

index.bs
miguelao: Also, we should discuss `pointsOfInterest` in the context of `MediaTrackConstraintSet`, this Section details what the UA returns as current settings, no constrain artifacts are needed here.

miguelao:
The only itchy point now is how to use a `sequence<Point2D>` as a `Constraint`, right @jan-ivar, @gmandyam? We could move on with this draft and address that concern as a separate issue if you guys are OK with it (note that this is an Editor's Draft anyway).

miguelao:
ping @jan-ivar

jan-ivar:
@miguelao I think at minimum we'd need to allow the `exact` and `ideal` keywords, since the constraints algorithm relies on these. E.g.:

```js
dictionary ConstrainPoint2DParameters {
    sequence<Point2D> exact;
    sequence<Point2D> ideal;
};
typedef (sequence<Point2D> or ConstrainPoint2DParameters) ConstrainPoint2D;
```
should do it.

miguelao:
Just added `ConstrainPoint2D` as per your suggestion, @jan-ivar, ready to go?

jan-ivar:
I'd still like @burnburn's opinion on the use of ` MediaSettingsRange` in `MediaTrackCapabilities`, as that might be controversial, but we can always discuss that in a follow-up.

miguelao:
\o/
@gmandyam, @alvestrand any further comments?

jan-ivar:
(Also, strictly speaking, I suppose MeteringMode should be DOMString in MediaTrackCapabilities and MediaTrackSettings, although since they're only ever returned from browser APIs and never used as inputs anywhere, it may not matter.)

miguelao:
Re. `MeteringMode` in `MediaTrackSettings` and `MediaTrackCapabilities` not being `DOMString`, I wanted to leave it as a strong type, but since it's the UA that returns these, I guess is not that crucial; I'm looking at you guys for platform consistency here. We can change it later on down the line if some else chimes in with a stronger vision.

jan-ivar:
@miguelao Consistency would be DOMString. i.e. follow how [facingMode](https://w3c.github.io/mediacapture-main/getusermedia.html#def-constraint-facingMode) is defined in the dictionary triad.

alvestrand:
If I remember rightly: The reason all constrainable types are DOMString not enum is that enums behave badly when extended - new values will throw exceptions (or rejected promises) on old browsers, which means that it's quite hard for an app to add values without extensive probing to figure out which values are acceptable. The WG decided that it would prefer to have DOMString so that any value can be passed, but the UA can deal with them appropriately (ignore ideal constraints it doesn't understand the value of, and return "impossible to satisfy" when those are encountered as "exact" values).

Yep, consistency is nice.

miguelao:
Done changing `MeteringMode` to `DOMstring` in `MediaTrackSettings` and `MediaTrackCapabilities` with the limitation of its values in prose. PTAL

miguelao:
@gmandyam, @alvestrand does this lgty?

gmandyam:
I looked through this and it seems to be good.  Thanks for taking this initiative @miguelao

miguelao:
I'm going to commit this PR given the previous responses: @alvestrand, @burnburn please have a reading when landed and shout if something smells pungent or can be improved.

ashlyrclark (commented):
diff

Amjidkhan6061:
> diff

I

PR #155, opened 2017-03-24
Update index.bs

gmandyam:
Add security and privacy section


gmandyam:
Marked as non-substantive for IPR from ash-nazg.

Issue #156, opened 2017-03-28
redEyeReduction as a sequence<boolean> is strange


domenic:
Why not just an enum, e.g. "never", "always", "controllable" or similar?

miguelao:
During the change to Constrainable Pattern we also updated this [capability](https://w3c.github.io/mediacapture-image/#dom-photocapabilities-redeyereduction), it reads now:
> If the source cannot do red eye reduction a single false is reported. If red eye reduction cannot be turned off, a single true is reported. If the script can control the feature, the source reports a list with both true and false as possible values.

Which is similar to what's done for [echoCancellation](https://w3c.github.io/mediacapture-main/#dom-mediatrackcapabilities-echocancellation) capability in medicapture-streams Spec and was timely suggested by @jan-ivar.

domenic:
Yes, I was saying that how it reads now is very strange, instead of using an enum. I guess if the idea is to align with another very strange API maybe that is worthwhile, but I am still not sure it is the right tradeoff.

jan-ivar:
@miguelao Actually, since `redEyeReduction` is a photoCapability, not a [Constrainable-Pattern](https://w3c.github.io/mediacapture-main/#capabilities)-using MediaTrackCapability, there's no technical reason it can't use enums or any API it wants. In fact, being different may be a benefit, to avoid confusing the two APIs. Sorry I missed this in review.

miguelao:
@jan-ivar what do you propose? An enum saying e.g.
```
enum redEyeReductionCapability {
  "unsupported",
  "fixed_false",
  "fixed_true",
  "configurable"
};
```
Or some such?

jan-ivar:
@miguelao What's the difference between `"unsupported"` and `"fixed_false"`?

I liked @domenic's `"never"`, `"always"`, `"controllable"`.

miguelao:
`"never"`, `"always"`, `"controllable"` it is! 👍

miguelao:
Closing by PR #165 .

PR #164, opened 2017-03-28
Made PhotoCapabilities.fillLightMode a sequence<FillLightMode>...

miguelao:
This PR:
- Makes `PhotoCapabilities.fillLightMode` a `sequence<FillLightMode>`, 
- removes `unavailable` from `FillLightMode`, since the condition of not-having-flash is encoded in returning an empty sequence and it doesn't make sense to use said `unavailable` for `PhotoSettings`,
- updates some text

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -179,10 +179,10 @@ interface ImageCapture {
 
 <pre class="idl">
   interface PhotoCapabilities {
-    readonly attribute FillLightMode      fillLightMode;
-    readonly attribute MediaSettingsRange imageHeight;
-    readonly attribute MediaSettingsRange imageWidth;
-    readonly attribute sequence&lt;boolean>  redEyeReduction;
+    readonly attribute sequence&lt;boolean>       redEyeReduction;
+    readonly attribute MediaSettingsRange      imageHeight;
+    readonly attribute MediaSettingsRange      imageWidth;
+    readonly attribute sequence&lt;FillLightMode> fillLightMode;
   };
 </pre>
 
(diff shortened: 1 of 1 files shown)


jan-ivar:
@miguelao This seems counter to https://github.com/w3c/mediacapture-image/issues/156, should we resolve that one first?

miguelao:
#156 is independent: in this PR I'm just moving the the idl to match the definitions below. We can talk about `redEyeReduction` separately :-)

index.bs
```diff
@@ -179,10 +179,10 @@ interface ImageCapture {
...
 <pre class="idl">
   interface PhotoCapabilities {
-    readonly attribute FillLightMode      fillLightMode;
-    readonly attribute MediaSettingsRange imageHeight;
-    readonly attribute MediaSettingsRange imageWidth;
-    readonly attribute sequence&lt;boolean>  redEyeReduction;
+    readonly attribute sequence&lt;boolean>       redEyeReduction;
+    readonly attribute MediaSettingsRange      imageHeight;
+    readonly attribute MediaSettingsRange      imageWidth;
+    readonly attribute sequence&lt;FillLightMode> fillLightMode;
```
jan-ivar: Actually, a sequence [cannot be an attribute](https://heycam.github.io/webidl/#idl-attributes). It's invalid WebIDL. We'd need:
```js
readonly attribute FrozenArray<boolean>    redEyeReduction;
readonly attribute FrozenArray<boolean>    fillLightMode;
```

miguelao: Changed in the latest commit.

Issue #169, opened 2017-03-31
getCapabilities needs to be async


beaufortfrancois:
As `MediaTrackCapabilities` can't be retrieved until a while after opening the device, it would be better if `getCapabilities` returns a promise so that it is always filled:

```js
const capabilities = await track.getCapabilities();
```

miguelao:
Note that [`getCapabilities()`](https://w3c.github.io/mediacapture-main/#dom-constrainablepattern-getcapabilities()) is part of the `mediacapture-streams` specification and that here we are limited by it.

That said, probably the idea in `mediacapture-streams` was to collect the capabilities once and only upon [`MediaDevices.getUserMedia()`](https://w3c.github.io/mediacapture-main/#mediadevices-interface-extensions) which is Promise-based:

`Promise<MediaStream>           getUserMedia(optional MediaStreamConstraints constraints);`

but, since in Chromium only the non-Promise based is implemented, we're in this situation.

alvestrand:
getUserMedia() is async in all its versions.

beaufortfrancois:
@alvestrand Good news.
@miguelao Maybe that means we can fix that then.

beaufortfrancois:
@miguelao (gentle ping)

miguelao:
`getCapabilities()` is innocent - Chrome implementation shouldn't need the timeout, but it needs it ATM -- let's track it in a Chromium Issue tracker, I'll add one to the B[link>ImageCapture component](https://bugs.chromium.org/p/chromium/issues/list?can=2&q=component%3ABlink%3EImageCapture+&sort=-modified&colspec=ID+Pri+M+Stars+ReleaseBlock+Component+Status+Owner+Summary+OS+Modified&x=m&y=releaseblock&cells=ids)

beaufortfrancois:
For future reference, here's the chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=711524

yellowdoge:
Since this applies to the Cr implementation only, and there's a tracking bug there (https://crbug.com/711524), let's close this bug.

Issue #148, opened 2017-01-26
Update examples if and when using Constrainable Pattern modus operandi.


yellowdoge:
Update/expand the examples if and when #124 is addressed.

PR #170, opened 2017-03-31
Corrected/updated example 1 (#148)

miguelao:
@beaufortfrancois FYI, corrected as per discussion elsewhere.  Also updated Codepen.


miguelao:
Also after https://github.com/GoogleChrome/samples/pull/503.

beaufortfrancois:
In Chromium current situation, we need a small timeout before calling `getCapabilities`, example should reflect this if that's really needed. If not, let's fix https://github.com/w3c/mediacapture-image/issues/169#issuecomment-290761692

PR #173, opened 2017-05-02
Correct header ID syntax for Security section.

scheib:


yellowdoge:
Well spotted! Thanks!

Issue #158, opened 2017-03-28
Provide API for getting / setting focus distance


alexshalamov:
The API allows to set manual focus mode, yet, there is no way to set focus distance manually. Maybe API can provide interface for getting focus range values as well as setting focus distance value.

dsanders11:
I'd like to bump this and voice my support for this as well. What is the current use case of setting `focusMode` to `manual` if you can't then provide a focus value? What focus value will be used, the last value before it was changed from auto to manual?

I have a strong use case for this API, and it could more or less be a drop-in replacement if a manual focus value is added to the API. Without a way to set a manual focus value it is not very useful as I have a use case where the user calibrates the camera settings once (sets zoom and focus value) and those settings are reloaded every time the page is used to provide a consistent experience as far as zoom and focus go. Currently this is accomplished in a hacky way by using WebSockets to talk to a server on the local machine which uses low-level APIs to set these values. I know that UVC and Windows APIs both have support for setting a manual focus value. The `android.hardware.camera2` API appears to have support for setting a focus distance, it's unclear if the older `android.hardware.Camera` API does.

@miguelao, anything I can do to help add this functionality to the spec?

yellowdoge:
Closing since the conversation has moved to #176 and its associated PR #175.

PR #175, opened 2017-05-16
Add focusDistance constraint

dsanders11:


index.bs
```diff
@@ -894,6 +912,65 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
   </pre>
 </div>
 
+## Update camera focus distance and {{takePhoto()}} ## {#example4}
+
```
yellowdoge: I'd love to have a codepen/jsfiddle/anything (a la [1]) like the other examples, so casual visitors could just click through and see this in action. 


[1] https://codepen.io/miguelao/pen/mAzXpN/left?editors=1010

yellowdoge:
I'd love @jan-ivar to have a reading as well.

yellowdoge:
Also @alexshalamov since #158 showed interest in this functionality.

index.bs
dsanders11: Perfect, I can add this to the PR. I don't have a codepen and didn't want to create one and mix it in when you'd already provided a few links under the same username.

index.bs
```diff
@@ -894,6 +912,65 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
...
+        // Map focus distance to a slider element.
+        const input = document.querySelector('input[type="range"]');
+        input.min = capabilities.focusDistance.min;
+        input.max = capabilities.focusDistance.max;
+        input.step = capabilities.focusDistance.step;
+        input.value = track.getSettings().focusDistance;
+
+        input.oninput = function(event) {
+          track.applyConstraints({
+            advanced : [{focusMode: "manual", focusDistance: event.target.value}]
```
jan-ivar: Please don't use `advanced`. Just do:
```js
track.applyConstraints({focusMode: "manual", focusDistance: event.target.value});
```

index.bs
```diff
@@ -894,6 +912,65 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
...
+    &lt;body>
+    &lt;video autoplay>&lt;/video>
+    &lt;img>
+    &lt;input type="range" hidden>
+    &lt;script>
+      var imageCapture;
+
+      navigator.mediaDevices.getUserMedia({video: true})
+        .then(gotMedia)
+        .catch(err => console.error('getUserMedia() failed: ', err));
```
jan-ivar: All errors end up here, not just "getUserMedia()" errors, so the string is misleading. More obvious if `gotMedia` were inlined.

index.bs
```diff
@@ -670,6 +686,8 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
     </div>
     </li>
 
+    <li><dfn>Focus distance</dfn> is a numeric camera setting that controls the focus distance of the lens.  The setting usually represents distance in millimeters to the optimal focus distance.</li>
```
jan-ivar: With `double` there's always the option to go unit-less (which I think we've done elsewhere here? but it escapes me). However, I suppose it's best to use what users expect, which seems to be millimeters if this is focal length, and my brief internet search wasn't way off. I know the TAG [recommends units like milliseconds](https://w3ctag.github.io/design-principles/#milliseconds) for time, but doesn't say anything about distance. Fyi.

dsanders11:
@jan-ivar, regarding the focal length comment, if you notice in the spec zoom is already listed as `Zoom is a numeric camera setting that controls the focal length of the lens.`.

The Windows API, Android API, and UVC spec all call it distance not length. You can find the API links in issue #176.

index.bs
dsanders11: I don't think there's anything that insists its millimeters, just that it is a logical measurement to use. I believe manufacturers can make it arbitrary if they'd like. The Android API doesn't mention units at all, just the term distance.

index.bs
dsanders11: The example is a cut and paste and modify from example 1 in the current spec, so both comments, regarding `advanced` and the error catching apply to example 1 as well. I'm happy to make the change, just a note that it should be consistent with the other examples.

yellowdoge: `advanced` is necessary bc of the way it's implemented in Chromium, and since for the examples sake the important thing is that they work, I'd say add a note here and please some one file a https://crbug.com/new ?

index.bs
jan-ivar: Do we know what units the Android API uses? Whatever we choose we should be specific about it, and not leave it up to devices. This PR says milliliters which sounds fine to me, I'm just not a domain expert.

dsanders11: Well, the PR says `usually represents distance` similar to several other spots in the current spec which say usually. I don't think we can say anything definitive because I don't believe anything says manufacturers have to tie it to anything at all, it can be completely arbitrary.

From the [Android API docs](https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics.html#LENS_INFO_FOCUS_DISTANCE_CALIBRATION):

> APPROXIMATE and CALIBRATED devices report the focus metadata in units of diopters (1/meter), so 0.0f represents focusing at infinity, and increasing positive numbers represent focusing closer and closer to the camera device. The focus distance control also uses diopters on these devices.
> **UNCALIBRATED devices do not use units that are directly comparable to any real physical measurement**, but 0.0f still represents farthest focus, and android.lens.info.minimumFocusDistance represents the nearest focus the device can achieve.

The Windows API clearly states millimeters. So, given the 3 different sources of info, Android says an uncalibrated device will use arbitrary units, a calibrated device will use diopters, and on Windows its millimeters (maybe).

alexshalamov: SI unit for distance is meters, not millimeters. Also, most of the lenses have clear markings (m or ft). Android uses diopters, which is not a problem, as it can be converted easily to meters.

dsanders11:
I've changed it to say meters as per @alexshalamov's comment.

Implementation notes being that for Windows/Linux the value will need to be converted from millimeters to meters, and for Android from diopters to meters, unless the device is `UNCALIBRATED` in which case the units are arbitrary and should probably be passed through unchanged.

index.bs
jan-ivar: @yellowdoge Limitations in an implementation should not limit what we put in spec examples IMHO. Specs drive implementations, not the other way around. I.e. our target audience is implementers. For instance, the [WebRTC examples](http://w3c.github.io/webrtc-pc/#advanced-peer-to-peer-example-with-warm-up) contain code no-one has implemented yet. Web developers will hopefully have blogs to rely on for things that work today.

`advanced` [should be avoided](https://github.com/w3c/mediacapture-main/issues/455) where possible as its semantic complexity is higher than needed.

index.bs
yellowdoge: Either way it's fine, we can add the example afterwards too.

index.bs
jan-ivar: > I figured platform/device specific stuff like that should be left alone since the device reports its possible values anyway.

@dsanders11 Valid point. Though having no conversion seems unfortunate in two respects:
 * Inability to show units in the UI.
 * Inability to hotswap one camera for another, and expect settings to persist.

I suppose apps could write javascript to transpose a setting's relative position in the legal range to the corresponding setting for a device with a different range. Though I'd rather we give this problem to the UAs.

dsanders11: @jan-ivar, fair points, especially about persisting settings across different cameras.

@yellowdoge, should I drop 'usually' and instead state that focus distance is in meters? Sounds like the consensus is to lock down the unit to meters, and if we're going to do that then it probably does more harm than good to say 'usually' since that leaves room for interpretation.

index.bs
dsanders11: I'll leave things as is for now, sounds like there's still some discussion on the content of the examples which may cause them to be revised in the near future anyway.

index.bs
jan-ivar: According to the mediacapture spec, we [should](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constrainablepattern-getcapabilities())  normalize hardware settings.

index.bs
yellowdoge: I see where you're coming from, but in this case I'd rather avoid a casual reader copy-pasting the example and not working: for that I'd add a comment in the code or a pop-up note in this Section (are examples normative?) detailing what's happening.

dsanders11:
@yellowdoge, how are we looking? I think I've addressed any concerns outside of the more general discussion on the contents of the examples. I think the only thing I still have unanswered is whether the wording should be changed to remove 'usually' from the description that says focus distance is in meters, so that it's more definitive.

yellowdoge:
Oops, sorry, I thought I had already approved this.  Just one thing though: could you check your [ipr status](https://labs.w3.org/hatchery/repo-manager/pr/id/w3c/mediacapture-image/175) (seems to be failing now).

dsanders11:
@yellowdoge, I'm waiting on W3C regarding that, I've been in contact over email with @dontcallmedom, so just waiting for that to come to a conclusion.

dsanders11:
@yellowdoge, still waiting on @dontcallmedom, who had an out-of-office email which said he'd be out until the 27th, so hopefully he's back tomorrow. I'll rebase once the IPR stuff is sorted.

yellowdoge:
@dsanders11  I see this PR has passed all the checks, does it mean your ipr status has been cleared? If so we should merge this PR.

dsanders11:
@yellowdoge, I noticed that as well, but I think it's actually that the repo has stopped doing the IPR status check. It says '1 successful check'.

IPR status is still pending, it required sending a document by snail mail to France, which was sent last week. Once that is received...there are more instructions. So, it might still be another week or two. 😕

dontcallmedom:
we've just received the snail mail; so while there will be a bit more of administrativia for @dsanders11 and me to go through, and while the IP manager will probably keep complaining, I feel confident enough that this can be safely merged form an IPR perspective.

yellowdoge:
Merging then !

PR #178, opened 2017-05-19
Make implementation-status.md link point to master branch

yellowdoge:
(and not to `gh-pages`, since there it isn't updated)


yellowdoge:
(@dsanders11 fyi, this PR should not conflict with #175 though)

Issue #177, opened 2017-05-16
Pan and Tilt Constraints


dsanders11:
What are the feelings on adding a pan constraint and a tilt constraint?

Both settings are part of the UVC spec. They're supported on Windows ([getRange_Pan](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376305(v=vs.85).aspx) and [getRange_Tilt](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376309(v=vs.85).aspx)) and via V4L2 on Linux (`V4L2_CID_PAN_ABSOLUTE` and `V4L2_CID_TILT_ABSOLUTE`).

They do not, however, appear to be supported on Android via either API (`android.hardware.camera` or `android.hardware.camera2`). Most built-in cameras on Android devices likely don't support them anyway, from my experience.

Logitech webcams support pan and tilt and it would be a great resource to have in the spec for fully controlling these types of cameras.

dsanders11:
@yellowdoge, thoughts? I do see in [this commit](https://chromium.googlesource.com/chromium/src/media/+/41dba5eecb31f0d2b040623f6fa8dafd4f102384) you mentioned "Avoid reset PAN/TILT controls due to Image Capture API not giving access
to them (yet)" so it sounds like you're in favor of adding the controls to the spec.

Anything learned there with V4L2 that makes the implementation seem too tricky to tackle at the moment?

yellowdoge:
Yeah, there's a private webcam extension ([code](https://cs.chromium.org/chromium/src/extensions/browser/api/webcam_private/?q=extensions/browser/api/webcam_private&sq=package:chromium&dr)), essentially a JS-surfaced IDL directly touching the filesystem `/dev/videoX` to manipulate zoom, pan and tilt for certain cameras ([more code](https://cs.chromium.org/chromium/src/extensions/browser/api/webcam_private/v4l2_webcam.cc?sq=package:chromium&dr&l=18)).   The issue that prompted that commit was to avoid resetting those controls because the clients of this extension API relied on the V4L2 API persisting the configuration values.

dsanders11:
Would that pose a large problem for adding pan and tilt constraints, or does it seem possible to work around?

yellowdoge:
[...]

Issue #126, opened 2016-12-07
Discussion of "zoom" does not consider mechanical zoom timing


alvestrand:
With non-electronic zoom, moving from one zoom setting to another takes time (seconds).

This specification should say how this is considered.
Options include:

- The desired zoom is what's returned in getPhotoCapabilities. The camera does what it wants.
- getPhotoCapabilities returns the current zoom. The user has to poll getPhotoCapabilities until the camera's zoom is at the desired point (and keep track of what the desired point is).
- setOptions() doesn't resolve its promise until the correct zoom has been reached

Either is viable, but which one is intended has to be specified.

miguelao:
Tentatively closing this issue, please re-read and comment or reopen.

alvestrand:
Does the promise resolve only once

alvestrand:
the note

> NOTE
If the UA can successfully apply the settings, the effect may be reflected, if visible at all, in videoStreamTrack. The result of applying some of the settings may force the latter to not satisfy its constraints (e.g. the frame rate).

is not reasonable; it specifies a number of dimensions in which the result of the call can vary, without giving normative guidance to the implementor on what should happen.

I'm particularly worried about the "may cause the track to be overconstrained" language - this will cause the track to go black for all other purposes.

Moving setOptions to be a parameter on takePhoto would resolve this issue, since takePhoto is already specified to be able to suspend the track and restore it to its original settings after the photo has been taken.

miguelao:
#124 has a discussion that should cover solving this issue.

PR #182, opened 2017-06-12
Add pan and tilt constraints

dsanders11:
As promised in issue #177.

Only real thing I can think of worth debating is whether to have the setting value be degrees, or arc seconds. They're trivial to convert between, and in UVC it is defined as arc seconds, as well as in more of the Windows APIs.

Degrees would be more intuitive for users, but since the underlying API will be returning the `max`/`min`/`step` in arc seconds, the `MediaSettingsRange` may look quite odd if you convert it to degrees, like from `-38.23` to `38.23` with steps of `0.453`. That could also lead to rounding problems in the conversion if the `step` in arc seconds is say 1000, that would be 0.277778 degrees, which is icky to deal with. As long as the camera "snaps" to the nearest valid step the rounding problems matter less, although it's never a great idea to have the value returned be different than the one you just set if the set succeeds. It's unintuitive and could lead to user code failing. Since either setting could be mechanical it may not apply immediately, and any end user code which keeps checking to see if it has changed to the correct value might fail or infinite loop if they set `21.234` degrees and that snaps to `21.2335` or something similar, since a simple `==` check would not be true.

@yellowdoge, I'm fairly confident these aren't available in the Android API. Would it make sense to use an `✕` in the implementation status to indicate that feature will never be available? The blank space would indicate that the feature is just not implemented yet, but could be in the future, versus `✕` means it will never be available.

### DIFF
--- a/implementation-status.md
+++ b/implementation-status.md
@@ -51,9 +51,11 @@ Feature/Platform          | Android                               | Linux/Chrome
 └ focusDistance           |   ([732807](https://crbug.com/732807))|                |                                         |     |
 └ focusMode               | ✓                                     | ✓              | 60 ([657128](https://crbug.com/657128)) |     |
 └ iso                     | ✓                                     |                |                                         |     |
+└ pan                     |                                       |                |                                         |     |
 └ pointsOfInterest        | ✓                                     |                |                                         |     |
 └ saturation              |                                       | ✓              | 60 ([657128](https://crbug.com/657128)) |     |
 └ sharpness               |                                       | ✓              | 60 ([657128](https://crbug.com/657128)) |     |
+└ tilt                    |                                       |                |                                         |     |
 └ whiteBalanceMode        | ✓                                     | ✓              | 60 ([657128](https://crbug.com/657128)) |     |
 └ zoom                    | ✓                                     | ✓              | 60 ([657128](https://crbug.com/657128)) |     |
 
--- a/index.bs
+++ b/index.bs
@@ -356,9 +356,11 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
 
     boolean brightness = true;
     boolean contrast = true;
+    boolean pan = true;
     boolean saturation = true;
     boolean sharpness = true;
     boolean focusDistance = true;
+    boolean tilt = true;
     boolean zoom = true;
     boolean torch = true;
   };
@@ -394,6 +396,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>contrast</code></dfn></dt>
   <dd>Whether <a>contrast</a> constraining is recognized.</dd>
 
+  <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>pan</code></dfn></dt>
+  <dd>Whether <a>pan</a> constraining is recognized.</dd>
+
   <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>saturation</code></dfn></dt>
   <dd>Whether <a>saturation</a> constraining is recognized.</dd>
 
@@ -402,6 +407,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
 
   <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>focusDistance</code></dfn></dt>
   <dd>Whether <a>focus distance</a> constraining is recognized.</dd>
+
+  <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>tilt</code></dfn></dt>
+  <dd>Whether <a>tilt</a> constraining is recognized.</dd>
 
   <dt><dfn dict-member for="MediaTrackSupportedConstraints"><code>zoom</code></dfn></dt>
   <dd>Whether configuration of the <a>zoom</a> level is recognized.</dd>
@@ -430,6 +438,8 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
     MediaSettingsRange   sharpness;
 
     MediaSettingsRange   focusDistance;
+    MediaSettingsRange   pan;
+    MediaSettingsRange   tilt;
     MediaSettingsRange   zoom;
 
     boolean              torch;
@@ -463,6 +473,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>contrast</code></dfn></dt>
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
+  <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
+  <dd>This reflects the <a>pan</a> value range supported by the UA.</dd>
+
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>saturation</code></dfn></dt>
   <dd>This reflects the permitted range of <a>saturation</a> setting. Values are numeric. Increasing values indicate increasing saturation.</dd>
 
@@ -471,6 +484,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>focusDistance</code></dfn></dt>
   <dd>This reflects the <a>focus distance</a> value range supported by the UA.</dd>
+
+  <dt><dfn dict-member for="MediaTrackCapabilities"><code>tilt</code></dfn></dt>
+  <dd>This reflects the <a>tilt</a> value range supported by the UA.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>zoom</code></dfn></dt>
   <dd>This reflects the <a>zoom</a> value range supported by the UA.</dd>
@@ -504,6 +520,8 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
     ConstrainDouble    sharpness;
 
     ConstrainDouble    focusDistance;
+    ConstrainDouble    pan;
+    ConstrainDouble    tilt;
     ConstrainDouble    zoom;
 
     ConstrainBoolean   torch;
@@ -540,6 +558,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>contrast</code></dfn></dt>
   <dd>See <a>contrast</a> constrainable property.</dd>
 
+  <dt><dfn dict-member for="MediaTrackConstraintSet"><code>pan</code></dfn></dt>
+  <dd>See <a>pan</a> constrainable property.</dd>
+
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>saturation</code></dfn></dt>
   <dd>See <a>saturation</a> constrainable property.</dd>
 
@@ -548,6 +569,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>focusDistance</code></dfn></dt>
   <dd>See <a>focus distance</a> constrainable property.</dd>
+
+  <dt><dfn dict-member for="MediaTrackConstraintSet"><code>tilt</code></dfn></dt>
+  <dd>See <a>tilt</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>zoom</code></dfn></dt>
   <dd>See <a>zoom</a> constrainable property.</dd>
@@ -577,6 +601,8 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
     double            sharpness;
 
     double            focusDistance;
+    double            pan;
+    double            tilt;
     double            zoom;
 
     boolean           torch;
@@ -613,6 +639,9 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
   <dt><dfn dict-member for="MediaTrackSettings"><code>contrast</code></dfn></dt>
   <dd>This reflects the current <a>contrast</a> setting of the camera.</dd>
 
+  <dt><dfn dict-member for="MediaTrackSettings"><code>pan</code></dfn></dt>
+  <dd>This reflects the current <a>pan</a> setting of the camera.</dd>
+
   <dt><dfn dict-member for="MediaTrackSettings"><code>saturation</code></dfn></dt>
   <dd>This reflects the current <a>saturation</a> setting of the camera.</dd>
 
@@ -621,6 +650,9 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
 
   <dt><dfn dict-member for="MediaTrackSettings"><code>focusDistance</code></dfn></dt>
   <dd>This reflects the current <a>focus distance</a> setting of the camera.</dd>
+
+  <dt><dfn dict-member for="MediaTrackSettings"><code>tilt</code></dfn></dt>
+  <dd>This reflects the current <a>tilt</a> setting of the camera.</dd>
 
   <dt><dfn dict-member for="MediaTrackSettings"><code>zoom</code></dfn></dt>
   <dd>This reflects the current <a>zoom</a> setting of the camera.</dd>
@@ -707,6 +739,14 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
     </li>
 
     <li><dfn>Focus distance</dfn> is a numeric camera setting that controls the focus distance of the lens.  The setting usually represents distance in meters to the optimal focus distance.</li>
+
+    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera.  The setting represents pan in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.</li>
+
+    <li><dfn>Tilt</dfn> is a numeric camera setting that controls the tilt of the camera.  The setting represents tilt in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values tilt the camera upward when viewed from the front, and negative values tilt the camera downward as viewed from the front.
+
+    <div class="note">
+      There is no defined order when applying <a>pan</a> and <a>tilt</a>, the UA is allowed to apply them in any order.  In practice this should not matter since these values are absolute, so order will not affect the final position.  However, if applying pan and tilt is slow enough, the order in which they are applied may be visually noticeable.
+    </div></li>
 
     <li><dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens.  The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1.  The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).</li>
 


yellowdoge:
Re. the `x`, I don't think it's needed: other platform columns, e.g. Mac or Linux, also have unexposed controls and we're just leaving those empty.

@anssiko, what's your opinion in the tilt/pan in arc seconds vs. degrees question ?

anssiko:
>@anssiko, what's your opinion in the tilt/pan in arc seconds vs. degrees question ?

I'll loop in @alexshalamov who's been looking at the spec more closely recently.

alexshalamov:
@anssiko I would vote for degrees. Most of the motorized heads or servo motors express rotation values in degrees (or radians). Easier to understand and use for web developers. Also, windows uses degrees for [pan](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376320(v=vs.85).aspx), [roll](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376325(v=vs.85).aspx) and [tilt](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376328(v=vs.85).aspx) for user space APIs and same units on the 'driver' level ([tilt](https://msdn.microsoft.com/library/windows/hardware/ff564454), [roll](https://msdn.microsoft.com/en-us/library/windows/hardware/ff564432(v=vs.85).aspx), [pan](https://msdn.microsoft.com/en-us/library/windows/hardware/ff564424(v=vs.85).aspx)).

If someone would like to use Image Capture API e.g., for astrophotography and HW supports fine step increments, 15 decimal places (double) should be enough to represent fractions of a degree / arc-second.

@dsanders11 BTW, why just Pan and Tilt without Roll? I have a camera where pan, tilt and roll can be controlled separately. Also, I think it would be better to specify order for applying 'rotation' constraints, so that vendors can implement interoperable API (pan-tilt-roll, zyx).

dsanders11:
> Easier to understand and use for web developers.

I agree degrees are more commonly used and intuitive, but I still worry about it leading to iffy code because of the rounding problems.

For example, a UI that lets the user control pan and tilt would probably want to wait for them to take effect before letting the user make another adjustment (or at least debounce their input) to prevent the camera from lagging behind and feeling out of sync. A developer might choose to implement that as:

```
input.oninput = function(event) {
  lockUI = true;  // Stand in for some method of locking the UI
  var panValue = event.target.value;
  track.applyConstraints({pan: panValue});

  var waitingForUpdate = true;

  while (waitingForUpdate) {
    if (track.getSettings().pan == panValue) {
      lockUI = false;
      waitingForUpdate = false;
    }
  }
}
```

Obviously that's not great code, but it's in the vein of code someone might write if they have to wait for a setting to take effect.

Now, if they had an input set up that used the pan value and step increments, with arc-degrees this would always be an integer and there is no rounding problem. With degrees, adding the step to the current value might lead to a very slightly different number then gets returned by `getSettings()`, which would break simple equality checking.

I may be overblowing the "risk" here, but I've definitely seen float rounding problems bite people in the real world, where order of operation leads to slightly different results. It's stuck with me enough to make me weary of situations where it could occur.

@yellowdoge, would the promise returned by `applyConstraints()` only resolve once the settings have taken effect (if pan takes 500ms to apply) or does it resolve once the constraint has been set, without waiting for it to take effect?

> @dsanders11 BTW, why just Pan and Tilt without Roll? I have a camera where pan, tilt and roll can be controlled separately.

Roll can easily be added, in my experience it isn't commonly available in consumer webcams so I didn't add it at this time. Most webcams in laptops and phones aren't going to have support for pan/tilt/roll at all, but I know Logitech brand webcams have support for pan/tilt, but not roll.

> Also, I think it would be better to specify order for applying 'rotation' constraints, so that vendors can implement interoperable API (pan-tilt-roll, zyx).

You'll have to expand on that a bit, I don't quite follow. There's not currently any notion of ordering constraints in this spec, I think adding it would change the API a decent amount unless you had an idea in mind?

yellowdoge:
> @yellowdoge, would the promise returned by applyConstraints() only resolve once the settings have taken effect (if pan takes 500ms to apply) or does it resolve once the constraint has been set, without waiting for it to take effect?

It would not wait, the Promise gets resolved upon passing the new settings successfully to the OS, and the user has to monitor, if desired, the correspondent setting.  A similar issue happens with `zoom` (@hta raised it in #126), if it's implemented by a mechanical device.

dsanders11:
Yea, so my main concern is that rounding issues with the floats might cause simple equality to not suffice, which would be an unfortunate side effect of using degrees.

Case in point, say the step value is 1000 arc-seconds. When dealing with degrees, that's:

```
step = 1000/3600;  // 0.2777777777777778
a = 45 + step*5;  // 46.388888888888886
b = 45 + step + step + step + step + step;  // 46.38888888888889

if (a != b) {
    alert("Here be dragons");
}
```

I guarantee that's going to mess with people's code. Anyone not familiar with the behavior of floats wouldn't have that kind of situation occur to them, and so they wouldn't check.

alexshalamov:
> Roll can easily be added, in my experience it isn't commonly available in consumer webcams so I didn't add it at this time. Most webcams in laptops and phones aren't going to have support for pan/tilt/roll at all, but I know Logitech brand webcams have support for pan/tilt, but not roll.

I'm not sure whether W3C APIs should be tailored for particular brand or type of device.

> You'll have to expand on that a bit, I don't quite follow.

When object is rotated in 3D space, order of rotations can affect end result http://quaternions.online/ , maybe I'm wrong :smiley: 

> I guarantee that's going to mess with people's code. Anyone not familiar with the behavior of floats wouldn't have that kind of situation occur to them, and so they wouldn't check.

I'm think it is not related to arcseconds or degrees, it is generic rounding problem. Spec doesn't say what happens when step has decimal places e.g., 1.89.

@yellowdoge Maybe specification could add note about rounding problem or introduce different `MediaSettingsRange` interfaces.

**For example:**
color temperature, iso, brightness, contrast, saturation, sharpness, zoom **=>** `unsigned long long`
focus, exposure comp **=>** `unrestricted double` (focus can be +Infinity)
tilt, pan, roll **=>** `long` (+-180deg)

@dsanders11 my personal opinion is that pan, tilt and roll are niche features and not directly related to photo / video capture capabilities. Those are the features of camera accesoires: motorized heads, 3-axis gimbals, conference webcam or security IP cams chassis. Looks like UVC and Windows APIs have those features for some legacy reasons, I'm not sure whether new specs should replicate same pattern.

Also, Windows is a good example of what not to do. Half of the methods [get / set arcseconds]( https://msdn.microsoft.com/en-us/library/windows/desktop/dd376341(v=vs.85).aspx), while other half [uses degrees](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376347(v=vs.85).aspx).

dsanders11:
> I'm not sure whether W3C APIs should be tailored for particular brand or type of device.

I'm not trying to tailor it to a particular brand or type of device, I was simply using a popular consumer brand as a way to gauge how common the roll control is in the wild.

> I think it is not related to arcseconds or degrees, it is generic rounding problem.

It's a generic problem with floats, but the conversion from arcseconds to degrees is what causes it to be more prevalent than it would be for other controls. Most other controls are integers, or have a couple decimal places. The conversion from arcseconds to degrees (dividing by 3600) leads to repeating decimals like the one in my example, which are likely to run into the rounding issue.

So while not specific to arcseconds, by their nature you're more likely to run into repeating decimals due to the conversion.

I won't put up too much of a fight here, I just think a spec should do its best to protect users from sharp edges if possible.

> @dsanders11 my personal opinion is that pan, tilt and roll are niche features and not directly related to photo / video capture capabilities. Those are the features of camera accesoires: motorized heads, 3-axis gimbals, conference webcam or security IP cams chassis. Looks like UVC and Windows APIs have those features for some legacy reasons, I'm not sure whether new specs should replicate same pattern.

I disagree with this assessment. They are not features limited to those listed camera accessories, they are features supported by many off the shelf webcams, of various brands, just not as common on cameras in tiny packages like phones or tablets. Roll is the more niche feature, which is why this PR is only for pan and tilt to begin with.

If you Google search [`"bmControls" pantilt`](https://www.google.com/search?q="bmControls"+pantilt) you can see some of the various webcams that support pan and tilt thanks to people posting the output of `lsusb` online. Webcams I've identified that support pan and tilt (but not roll) in about 5 minutes of looking:

* [vendorId: Acer (probably builtin laptop webcam)](https://mail.gnome.org/archives/ekiga-list/2015-April/msg00013.html)
* [HP Webcam HD 5210](http://buttersideup.com/files/hp-5210-lsusb.txt)
* [vendorId: Lenovo (probably builtin laptop webcam)](http://webcache.googleusercontent.com/search?q=cache:6uL8U1VpjSwJ:www.os2world.com/forum/index.php%3Faction%3Ddlattach%3Btopic%3D196.0%3Battach%3D228+&cd=7&hl=en&ct=clnk&gl=us)
* [Dell Inspiron 1520 builtin webcam](https://forums.virtualbox.org/viewtopic.php?f=2&t=3740)
* [Microsoft LifeCam HD-3000](https://6xq.net/webcam-firmware/hd3000_lsusb.txt)
* A dozen different models of Logitech webcams

I believe there to be broad support for pan and tilt among laptop integrated webcams and external USB webcams, without any sort of accessory required.

@dsanders11 my personal opinion is that pan, tilt and roll are niche features and not directly related to photo / video capture capabilities.

Pan and tilt, especially when paired with zoom, have good applicability for use with face tracking, eye tracking, and other such use cases.

alexshalamov:
@dsanders11 Thanks for info, those links basically prove that pan and tilt is a niche segment. I checked links for integrated cameras, looks like laptop manufacturers just soldered 3rd patty components and took drivers that report pan/tilt capability. I doubt cameras installed in a window would be able to pan and tilt properly.

Some external webcams or security cameras do have such capability and I mentioned that before. Do you have rough estimate for market share of external cameras having pan/tilt capabilities and all mobile devices + laptop with integrated cameras?

As for degrees vs arcseconds, it would be nice to wait for reply from @yellowdoge, in my opinion, using double for all settings might be not an ideal solution.

alexshalamov:
@dsanders11 Do you see any problems with using degrees vs arcseconds? For example, do external camera motors support less than 1 degree steps?

I forgot to mention that other w3c APIs use degrees to express rotation, DeviceOrientation, ScreenOrientation use degrees. OrientationSensor uses quaternion that can be converted to degrees, maybe it would be better to 'sync' units that express rotations across w3c apis.

dsanders11:
> @dsanders11 Thanks for info, those links basically prove that pan and tilt is a niche segment. I checked links for integrated cameras, looks like laptop manufacturers just soldered 3rd party components and took drivers that report pan/tilt capability.

Again, I disagree with this assessment. I don't see any evidence that they're misreporting having a pan and tilt functionality which they don't have. Look through the information in those links again, no two of them are reporting having the exact same set of controls, they're all slightly different. This isn't a case of generic drivers misreporting features.

> I doubt cameras installed in a window would be able to pan and tilt properly.

I think this is the root of different understandings. I believe when you hear pan and tilt you're imagining *mechanical* pan and tilt, with motors, of the source you mentioned in your description of accessories: "motorized heads, 3-axis gimbals, conference webcam or security IP cams chassis."

This is not the only way that cameras can provide the pan tilt functionality. Many of these cameras reporting pan and tilt have it as a feature of the CMOS chip used, like [this one here](http://www.onsemi.com/pub/Collateral/MT9M131-D.PDF). The datasheet for that CMOS chip (which is a tiny 4.6mm x 3.7mm) lists `"Electronic Pan, Tilt, and Zoom"` as a feature of the chip. Cameras using chips with *electronic* pan and tilt don't need anything mechanical inside their housing and simply have a lens over the CMOS chip. Their range of pan and tilt is much smaller, for example, 10 degrees in any direction, but it exists.

[Here on the Logitech site they cover controls for their C920 camera](http://support.logitech.com/en_us/article/Customize-HD-Pro-Webcam-C920-Settings-with-Logitech-Gaming-Software?product=a0qi00000069uigAAA). Of particular interest are these two bullets:

> * **Pan/Tilt controls** — Use the left, right, up, and down arrows to fine-tune the field of vision.
> * **Follow my face** — Click on the checkbox to allow the camera lens to automatically pan and tilt in order to follow your face.

[This is what a teardown of that webcam looks like](https://lieksa.pp.fi/oh7fes/cameras/focus.htm). There's nothing mechanical in there, the pan and tilt are being provided electronically.

Notice that the Logitech software has one of the features I was mentioning adding these controls to the spec would allow web developers to write - using the pan and tilt controls for face tracking.

I'll gladly concede the arc-degrees versus degrees in deference of other specs using degrees and if that's what others want. I will not, however, concede that pan and tilt are too niche to be in the spec.

alexshalamov:
> Many of these cameras reporting pan and tilt have it as a feature of the CMOS chip used, like this one here.

That sensor has max 1 degree rotation. Lets say, with 50-60 deg FOV for webcam, that would not be much of a pan / tilt.

> EDIT: Doing some more research, it's a bit of a jumble on what this is called in literature, but it's usually electronic or digital pan/tilt/zoom. Sometimes called ePTZ or digital PTZ.

If I remember correctly term digital PTZ is mostly used for security IP cams with ultra wide FOV. Stream from such cameras can be projected (perspective projection) to normal FOV and users can pan / tilt digitally.

@dsanders11 Do you think pan / tilt rotations need to be with less than 1 degree steps? Can `long` be used instead of `double`? If you think that roll can be added later, I think it is fine. In my opinion, roll can be important for leveling horizon when mechanical pan/tilt/roll is used, gimbals, drones, etc. BTW, did you check my question about pan tilt roll order? I still think it is important, that is why it is PanTilt or PanTiltRoll, standard Euler angles rotation order (ZYX, yaw-pitch-roll)

dsanders11:
> That sensor has max 1 degree rotation. Lets say, with 50-60 deg FOV for webcam, that would not be much of a pan / tilt.

Webcams I've dealt with that were providing electronic pan tilt were in the range of +10 to -10 degrees.

> @dsanders11 Do you think pan / tilt rotations need to be with less than 1 degree steps? Can long be used instead of double?

I think it needs support for less than 1 degree steps because there's nothing that says it can't be less than 1 degree. While in practice it may be uncommon, how would the API act if there is a camera with 0.5 degree steps? I suppose this spec could round step to the nearest degree, but that could again run into a camera that has something like 0.75 degree step which would mean a rounded degree value is often not a valid value.

The underlying APIs that are used for the camera control (UVC and Windows [getRange_Pan](https://msdn.microsoft.com/en-us/library/windows/desktop/dd376305(v=vs.85).aspx)) return the step value as arc degrees so I think we have to assume that it can be less than 1 degree.

>  BTW, did you check my question about pan tilt roll order? I still think it is important, that is why it is PanTilt or PanTiltRoll, standard Euler angles rotation order (ZYX, yaw-pitch-roll)

You'll have to make an example of how to accommodate this in the spec, I don't see how enforcing an order can easily be added to the current API.

Wouldn't order only matter if using relative pan tilt roll? So far everything in the spec (correct me if I'm wrong) is using *absolute* values for constraints, not *relative* values. While order certainly matters when making relative transformations, does it matter when the values are absolute? Doing pan 5 degrees, roll 5 degrees, pan -5 degrees, roll -5 degrees would not get you back to default position when using relative values, doing pan to 5 degrees, roll to 5 degrees, pan to 0 degrees, roll to 0 degrees will always get you back to default position since they're absolute and the last value set was 0.

yellowdoge:
To add my POV to this discussion, PanTilt is exposed by the [UVC 1.5 Spec](https://tinyurl.com/uvc-1-5-specification), Sec. "4.2.2.1.14 PanTilt (Absolute) Control" and "4.2.2.1.15 PanTilt (Relative) Control"; UVC is the underlying bus for all Desktop OSs, including Windows (whereas mobile platforms use MIPI). IMHO, that gives a large and stable enough deployed base to justify its inclusion in the Spec, also taking into account that if these are not supported by the concrete hardware, they won't appear in the `MediaStreamTrackCapabilities`.  Also, the closer the Spec looks like the UVC the easier they'll match; UVC defines the units as `Signed Number` of size `4` bytes, so: int?.

Re. `pantilt` order, I guess @alexshalamov is aiming at including, or not, the path to get from a given "coordinate" to another as part of the Spec, e.g. from `pan=0`, `tilt=0`, to `pan=1`, `tilt=1`, the underlying hardware could manipulate `pan` first, then `tilt`, or viceversa, or both at the same time, and, e.g. at what speed.  Looking at the UVC 1.5 Spec I don't see much definition for these issues, so we might simply need to explicit that the way "coordinates" in the `pantilt` space are reached is UA specific.  

Also note that if `roll` is included, it changes the game since the application of `pt` and `roll` would not commutative and we'd be in the deep waters of the quaternions.   Since `roll` is separated in the UVC spec, perhaps we should leave it for another PR, and then specify some order. Personally I haven't found any webcams with roll, so we might also need some evidence of implementations in the wild.

dsanders11:
> Looking at the UVC 1.5 Spec I don't see much definition for these issues, so we might simply need to explicit that the way "coordinates" in the pantilt space are reached is UA specific.

Well, more than that, the UVC 1.5 Spec defines the PanTilt as a single control, you set both in one USB command. While you could use the relative version to enforce an order, setting 0 for the relative movement of the non-target control, the ImageCapture spec here is using absolute values, and there's no way in the UVC 1.5 Spec to set the absolute pan and tilt values separately. You have to define them at the same time, so the order in which they are applied is entirely undefined in the spec. The implementation could perform the pan and tilt in any order when it has been set both at the same time.

> Personally I haven't found any webcams with roll, so we might also need some evidence of implementations in the wild.

You can find some [if you search it](https://www.google.com/search?q="bmControls"+roll) particularly on the second page but it seems like they're less common than webcams with pan tilt.

yellowdoge:
>  so the order in which they are applied is entirely undefined in the spec.

Precisely. Could you then add literature to this effect in the PR please? A `<div class="note">` should be enough.  
Also note that you have to rebase `implementation-status.md`.

dsanders11:
@yellowdoge, note added and I did a force push to the branch because something had gotten wonky and it was showing commits not in upstream even though it didn't have any effect on the changes in the branch. Should be back to a single clean commit.

The language in the note may be a bit wordy but was trying to convey what has been discussed here. With absolute values I still see no reason to worry about the ordering, but hopefully the note is enough to get this PR back on track.

yellowdoge:
Cool, thanks! @jan-ivar , @alexshalamov ptal.

jan-ivar:
@yellowdoge I think you meant @dsanders11. 😏

Reading through this, I don't have a strong opinion on rounding errors, but I would like to point out the several other ways that @dsanders11's [example](https://github.com/w3c/mediacapture-image/pull/182#issuecomment-308837309) won't work:
 1. Busy-spinning on `getSettings()` waiting for it to change will always hang, because implementations generally queue tasks to update JS state. Use `await`.
 2. `track.applyConstraints({pan})` is the same as `track.applyConstraints({pan: {ideal: pan}})` and does not guarantee a value. Your page may be one of several tabs accessing the singular device simultaneously, and the browser chooses the final value. Use `track.applyConstraints({pan: {exact: pan}})` to force a value or fail.
 3. Does `applyConstraints()` wait for the motor? I dunno. I've filed https://github.com/w3c/mediacapture-main/issues/470.

dsanders11:
> Reading through this, I don't have a strong opinion on rounding errors, but I would like to point out the several other ways that @dsanders11's example won't work

It also won't work because it simply sets a magical `lockUI` variable. It wasn't meant to be a working example (which is why I said it wasn't great code) but quick pseudo-code to highlight the potential problem lurking due to the float rounding.

It's not an example being inserted into the spec, I don't see how it is useful in moving this discussion forward to pick it apart on ancillary details.

The point being illustrated was that a seemingly intuitive use of simple equality may unexpectedly fail. You also used a simple equality check in the code you supplied in the issue you filed regarding `applyConstraints()`. You have to go out of your way to write code that would be immune to the float rounding problem. We have an opportunity to insulate end users from that problem, one many would not be aware of.

jan-ivar:
@dsanders11 Sure, my point was there were other assumptions here, not just that doubles match. The big issue is w3c/mediacapture-main#470 which questions whether `getSettings()` can be used like this at all. i.e. whether it returns an actual live hardware value, or just the last configured value (that the hardware is presumably moving towards asap). If it's the latter, then this is not a use-case showing that rounding is a problem. Not saying there aren't others. Just saying.

dsanders11:
@jan-ivar, I contend that even if `getSettings()` returns the last configured value immediately, this is still a use case highlighting the rounding problem.

As you pointed out, if you don't specify `exact` then the constraint may not actually be applied. How would one check if it had been applied? The promise returned by `applyConstraints` resolves to `void` if successful, so it doesn't provide you the information you need. A logical way to check would be to use `getSettings()` to see if your ideal settings had been set. For example, I may want: `{pan: {ideal: pan}, tilt: {ideal: tilt}, width: {exact: 640}, height: {exact: 480}}`.

If applying those constraints does not cause the promise to reject then I can be sure width and height were fine, but I don't know if pan or tilt took. So I check them with `getSettings()` and change the UI accordingly if they were not accepted, and we're back to the float rounding problem with simple equality. My check may say they weren't applied when in actuality they were, just my equality check failed due to the rounding.

Sure, this can be avoided by writing the code differently, such as applying each constraint one at a time with `exact` so that it is clear when a constraint fails. There's always ways around shortcomings, if you know the shortcomings.

If we go with degrees over arc-seconds then I'll likely add a note to the spec warning about these rounding problems and simple equality checks, although most casual developers are not going to be reading the spec itself.

jan-ivar:
FWIW, since the UA actually stores the constraints, could implementations fix these rounding errors?

E.g. If `track.getConstraints().pan.ideal` is `21.234`, then `track.getSettings().pan` could return `21.234` instead of `21.2335`.

Even if it changed back to `21.2335` once you removed the constraint, it wouldn't be that weird (who knows, maybe the UA decided to move it a tiny bit).

There seems to be some allowance for spec'ing this sort of thing in this mediacapture spec [note](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constrainablepattern-getcapabilities()).

dsanders11:
@jan-ivar, I don't think there's anything about being stateful in this spec. Your suggestion would require the implementation of the spec to keep a separate state, where as I believe current implementations just pass the values to the OS-level APIs and read them back accordingly. The problem with keeping state is that the state could be changed by something outside of the implementation, or simply another tab, which could lead to inconsistency.

The note you linked to, while interesting, is more about 1-to-1 mappings it would seem. The end also seems to suggest that `getSettings()` would still return the true value (although it doesn't say it has to).

I think that solution would slightly side-step the issue, but isn't an absolute fix, so I'm not sure it would be worth doing a half-measure if it adds extra complexity to implementations. The main issue is that order of operation when using `step` can cause rounding problems with floats, so while the mapping might save the programmer some of the time, a different path through the code may perform the math in a different order which would result in a different value, and we're back to square one.

jan-ivar:
> I believe current implementations just pass the values to the OS-level APIs and read them back accordingly

@dsanders11 That's too expensive because `track.getSettings()` is synchronous. In Firefox at least, we don't access devices on the main thread. Instead we queue a task on the main thread when state changes, to store it there, where JS can read it.

Regardless, even if we accessed it directly, the only state we'd need to round values before returning them is `track.getConstraints()` which is already stored state according to the spec.

Adding complexity to implementations is warranted if it removes complexity for users.

dsanders11:
@jan-ivar, if the spec stated that pan and tilt were in degrees with at most 2 decimal points, that could remove some of the sharp edge from the user side. It's unlikely anyone would ever need less than `0.01` of a degree, and users could do comparisons with simple JavaScript thanks to type coercion:

```
track.getSettings().pan == pan.toFixed(2)
```

There's still the sharp edge where they need to know to use `toFixed(2)`, but it's at least a simple and clean solution to the problem in JavaScript.

The only way to fully remove the sharp edge is to use arc-degrees, but it seems like there's not much support for that.

jan-ivar:
@dsanders11 Again, I don't have a strong opinion atm. Just noting that if people can remember that then they can probably remember:
```js
track.getSettings().pan.toFixed(2) == pan.toFixed(2)
```
which is actually less to remember because it's not critical they get the `2` right.

FWIW, [aspectRatio](https://w3c.github.io/mediacapture-main/getusermedia.html#def-constraint-aspect) is rounded to the 10th decimal point (though nobody implements that constraint yet AFAIK). The constraints language also has `min` and `max` which may be used over `exact`.

I don't know much about arc seconds.

dsanders11:
@jan-ivar, I'm not worried about people remembering as much as being aware in the first place. In the ideal world where everyone does a deep on the spec before coding then we wouldn't even have to worry about this as a note in the spec would solve it all. :-)

Having the spec specify rounding pan and tilt to a fixed 2 decimal places would have the benefit of making it more *naturally discoverable* by users. Where as the float rounding problem would be subtle and hard to debug since it would work fine most of the time, with 2 fixed decimal places the equality check failing would be more obvious to users as the number of decimal places in each would stick out.

I still don't love it because again most of the time it would lurk, if developing with a camera that only does 1 degree steps for pan and tilt you'd never notice the issue until someone with a camera that used a non-integer step tried it.

> FWIW, aspectRatio is rounded to the 10th decimal point (though nobody implements that constraint yet AFAIK).

Good info to have, that there's precedence for rounding to a decimal point with constraints. The main reason it's more prevalent with this spec is `MediaSettingsRange` encourages calculating values with `step` instead of fixed calculations.

> I don't know much about arc seconds.

AFAIK there's not much to know: an arc-second is 1/3600 of a degree. The applicability here is that none of the underlying OS APIs support less than 1 arc-second granularity so by using arc-seconds you can use integers to represent rotation and avoid the float rounding problem.

jan-ivar:
@dsanders11 From my POV I see two viable options:
 1. Use arc seconds, or
 2. Use degrees without rounding, but auto-snap values within 0.000001 of applied constraint.

Either would seem to me to solve most if not all rounding issues.

I don't like the `.toFixed(2)` solution, as it would seem to only exacerbate rounding errors with `step`.

Clearly, if arc seconds were commonly used, they're the inherently simpler option, but I don't have the domain knowledge to know if they are. Naively, I think I'd still pick degrees, but I'm not the target audience. I'd also follow any web precedent here if there is one.

yellowdoge:
This Pr seems to have gone quiet for a while now :-)
Any resolution, @dsanders11, @jan-ivar ?

yellowdoge:
@dsanders11 , @jan-ivar, friendly ping :-)

Note that in Chrome we already have a [private API](https://cs.chromium.org/chromium/src/extensions/common/api/webcam_private.idl?q=pan+tilt&dr=C&l=33) where the Range is not specified, i.e. developers would use whatever the driver says.

alvestrand:
How should this API relate to the WebUSB API?
Given that the PTZ controls mentioned so far are all USB controls, perhaps the driver should be stated in terms of USB primitives?

(BTW, the spec HAS to admit to the fact that PTZ is not instant - at any given moment, the position of the camera is pointing in a definite direction, but that direction may or may not be the target position.)

jan-ivar:
@yellowdoge I summarized my position [earlier](https://github.com/w3c/mediacapture-image/pull/182#issuecomment-313424967). I don't have any new info.

> perhaps the driver should be stated in terms of USB primitives?

@alvestrand What do you mean by "driver" and "stated" here?

> BTW, the spec HAS to admit to the fact that PTZ is not instant

We fixed this is in https://github.com/w3c/mediacapture-main/pull/512/files to say settings are *target values*, not measured. Is that normatively sufficient? A note might help of course.

riju:
Thanks @dsanders11 for the PR. As you rightly pointed out, both the UVC spec (section 4.2.2.1.14) as well as Windows getRange_Pan API use arc seconds as the unit for pan and tilt settings.

@jan-ivar summarized the [options for the spec](https://github.com/w3c/mediacapture-image/pull/182#issuecomment-313424967) succinctly and based on the discussion in this PR, related issue, and data provided it seems the option “1. Use arc seconds” is the option everyone can live with. As a further data point, I also reviewed some native applications like gucview, and observed they also use arc seconds as the unit.

Given pan and tilt are slightly niche features, my expectation is web developers who’ll make use of these features in their web apps are comfortable working with arc seconds.

I [implemented this feature in Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/1508519) ([video](https://github.com/riju/WebCamera/blob/master/panTilt.gif) - note: big file) according to this spec PR to try it out, and noticed no issues. I would like to hear if people have any concerns in merging this PR.

CC : @anssiko

yellowdoge:
@dsanders11 are we then OK with option “1. Use arc seconds” where everyone seems happy?

dsanders11:
@riju, @yellowdoge, apologies for the late response. Yes, arc-seconds is great by me.

PR #204, opened 2019-03-22
Fix implementation status

riju:
Adding the exposureTime and focusDistance implementation status.


yellowdoge:
https://crbug.com/823316 appears in crbug.com as started, can you mark it as fixed plz? Thanks!

riju:
Ofcourse. Done :)

PR #205, opened 2019-03-22
Add tag-security-privacy.md

### LINKED ISSUE #172: Fill Security & Privacy questionnaire #122
The spec just recently started a privacy considerations section; it would be good to go through the [W3C TAG self-review security & privacy questionnaire](https://w3ctag.github.io/security-questionnaire/) to help ensure we didn't miss any other obvious risk.

riju:
Add the tag-security-privacy document to assist in the TAG review.
This was requested in the [Intent to Ship thread](https://groups.google.com/a/chromium.org/d/msg/blink-dev/j-Q08QgBipM/RgGOQxG-BwAJ)

This document right now covers Pan/Tilt but I guess is applicable to
most of the other camera properties on this specification.

@yellowdoge : PTAL. A document like this is a pre-requisite for TAG Review Request.

Fixes #172

### DIFF
--- a/tag-security-privacy.md
+++ b/tag-security-privacy.md
@@ -1,0 +1,65 @@
+# TAG Security & Privacy Review Document
+
+Answering questions from [the questionaire doc here](https://w3ctag.github.io/security-questionnaire/).
+
+## Questions to Consider
+
+### 3.1. Does this specification deal with personally-identifiable information?
+Not directly, only if Pan/Tilt is supported by the camera and the value of Pan/Tilt.
+
+### 3.2. Does this specification deal with high-value data?
+No.
+
+### 3.3 Does this specification introduce new state for an origin that persists across browsing sessions?
+No.
+
+### 3.4. Does this specification expose persistent, cross-origin state to the web?
+No.
+
+### 3.5. Does this specification expose any other data to an origin that it doesn’t currently have access to?
+No.
+
+### 3.6. Does this specification enable new script execution/loading mechanisms?
+No.
+
+### 3.7. Does this specification allow an origin access to a user’s location?
+No.
+
+### 3.8. Does this specification allow an origin access to sensors on a user’s device?
+No.
+
+### 3.9. Does this specification allow an origin access to aspects of a user’s local computing environment?
+No.
+
+### 3.10. Does this specification allow an origin access to other devices?
+No.
+
+### 3.11. Does this specification allow an origin some measure of control over a user agent’s native UI?
+No.
+
+### 3.12. Does this specification expose temporary identifiers to the web?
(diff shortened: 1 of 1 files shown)


tag-security-privacy.md
```diff
@@ -0,0 +1,65 @@
+# TAG Security & Privacy Review Document
+
+Answering questions from [the questionaire doc here](https://w3ctag.github.io/security-questionnaire/).
+
+## Questions to Consider
+
+### 3.1. Does this specification deal with personally-identifiable information?
+Not directly, only if Pan/Tilt is supported by the camera and the value of Pan/Tilt.
```
yellowdoge: Wait, this new file affects the whole spec, so this should say something else than just about `pan` and `tilt`, right?

tag-security-privacy.md
```diff
@@ -0,0 +1,65 @@
...
+No. [GETUSERMEDIA](https://www.w3.org/TR/mediacapture-streams/) might expose deviceID. This specification does not expose anything extra.
+
+### 3.13. Does this specification distinguish between behavior in first-party and third-party contexts?
+No.
+
+### 3.14. How should this specification work in the context of a user agent’s "incognito" mode?
+The same as outside of incognito.
+
+### 3.15. Does this specification persist data to a user’s local device?
+No.
```
yellowdoge: I guess you're finding all these questions because this spec never started its way to CR, so this questionnaire was never run through and there's no security and privacy section in the Spec. With that said, I could see a malicious app setting e.g. `pan`, `tilt` and `zoom` e.g. to minimally different values and scooping them later on as an active fingerprinting technique, WDYT?

riju: > there's no security and privacy section in the Spec

The privacy and security considerations discussed in [GETUSERMEDIA](https://www.w3.org/TR/mediacapture-streams/) apply to this extension specification.

> I could see a malicious app setting e.g. pan, tilt and zoom e.g. to minimally different values and scooping them later on as an active fingerprinting technique, 

I think yes, and of course it applies to other constrainsts also. Spec might also add a note about implementation that - Resetting to a default value, say 0 Pan, 0 Tilt after every session.

tag-security-privacy.md
riju: That's true. We can update this by -> 

_Using the values of the properties exposed, attacker might narrow down the subset of cameras probably used by the user._ 

The point is if that is _valuable enough_ information and does that need "rounding" or coarse steps/ quantization to mitigate this. With Pan/Tilt, if camera allows steps of 1 arc-degrees, we might allow implementation to allow steps of say 10 arc-degrees for all cameras, although ranges may give some hint. IMHO, since getUserMedia() is gated by user activated permission, we should try to give whatever native application can give.

riju:
The questions in the TAG Security and Privacy document have changed, so this PR is now going to be superseded by PR #242 .

PR #218, opened 2020-03-20
Update pan, tilt and zoom constraints

eehakkin:
Our model of thinking has been :
The Pan-Tilt-Zoom permission can only be requested (and only shown in the prompt) if the currently connected/selected camera supports it.

Camera permission without PTZ (obtained using an older version of the user agent, or with another camera) is not implicitly upgraded to Camera permission with PTZ (even if the hardware supports it). The permission has to be re-requested.

In w3ctag/design-reviews#358, TAG wants to make sure that from the web app’s perspective: Pan-Tilt-Zoom needs to be explicitly requested as an extension to video capture permission.

To facilitate the previous requirements we want to add a new CameraDevicePermissionDescriptor dictionary introduced in w3c/permissions/pull/204 to differentiate between the two cases and to modify media capture so that using a pan, tilt or zoom constrainable property prompts for a `{name: "camera", panTiltZoom: true}` permission and that it is possible to request elevated permission prompt during `getUserMedia` call in order to avoid double permission prompt (`{name: "camera"}` during `getUserMedia` and `{name: "camera", panTiltZoom: true}` during `applyConstraints`).

This allows the following kind of code:
```
navigator.mediaDevices.getUserMedia({
    video: {
        height: { ideal:  720, min: 240, max: 1080 },
        width:  { ideal: 1280, min: 320, max: 1920 },
        // These empty constraints are to make getUserMedia
        // to request permission to use a permission descriptor
        // {name: "camera", deviceId: deviceId, panTiltZoom: true}
        // instead of normal {name: "camera", deviceId: deviceId}
        // so that subsequent applyConstraints call do not have to
        // prompt for  new permissions.
        pan:    true,  // or {}
        tilt:   {},    // or true
        zoom:   true   // or {}
    }
})
.then(async mediaStream => {
    const video = document.querySelector('video');
    video.srcObject = mediaStream;
    await sleep(1000);  // crbug.com/711524
    const track = mediaStream.getVideoTracks()[0];
    const capabilities = track.getCapabilities();
    const settings = track.getSettings();
    if ('pan' in capabilities) {
        const input = document.getElementById('pan-range');
        input.min = capabilities.pan.min;
        input.max = capabilities.pan.max;
        input.step = capabilities.pan.step;
        input.value = settings.pan;
        input.oninput = function(event) {
            // Applying pan (or tilt or zoom) constraint
            // requests permission to use a permission descriptor
            // {name: "camera", deviceId: deviceId, panTiltZoom: true}
            // which is already granted during successful getUserMedia
            // call above thus this applyConstraints call does not prompt for
            // any new permissions.
            track.applyConstraints({advanced: [{pan: event.target.value}]});
        };
        input.hidden = false;
    }
    if ('tilt' in capabilities) {
        // ditto
    }
    if ('zoom' in capabilities) {
        // ditto
    }
});
```

For API, it would of course be possible to combine `{video: {pan: true, tilt: true, zoom: true}}` to `{video: {panTiltZoom: true}}` but in that case one had to use `navigator.mediaDevices.getUserMedia({video: {panTiltZoom: true, zoom: {min: 2}}})` instead of `navigator.mediaDevices.getUserMedia({video: {zoom: {min: 2}}})` or similar if one wants to use pan, tilt or zoom constraints in getUserMedia. With separate pan, tilt and zoom (instead of a combined panTiltZoom) it is also easier to expand to roll if pan/tilt/roll/zoom cameras ever become more widespread. But please share your opinions.

/cc @riju

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -523,27 +523,27 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
 
 <pre class="idl">
   partial dictionary MediaTrackConstraintSet {
-    ConstrainDOMString whiteBalanceMode;
-    ConstrainDOMString exposureMode;
-    ConstrainDOMString focusMode;
-    ConstrainPoint2D   pointsOfInterest;
+    ConstrainDOMString           whiteBalanceMode;
+    ConstrainDOMString           exposureMode;
+    ConstrainDOMString           focusMode;
+    ConstrainPoint2D             pointsOfInterest;
 
-    ConstrainDouble    exposureCompensation;
-    ConstrainDouble    exposureTime;
-    ConstrainDouble    colorTemperature;
-    ConstrainDouble    iso;
+    ConstrainDouble              exposureCompensation;
+    ConstrainDouble              exposureTime;
+    ConstrainDouble              colorTemperature;
+    ConstrainDouble              iso;
 
-    ConstrainDouble    brightness;
-    ConstrainDouble    contrast;
-    ConstrainDouble    saturation;
-    ConstrainDouble    sharpness;
+    ConstrainDouble              brightness;
+    ConstrainDouble              contrast;
+    ConstrainDouble              saturation;
+    ConstrainDouble              sharpness;
 
-    ConstrainDouble    focusDistance;
-    ConstrainDouble    pan;
-    ConstrainDouble    tilt;
-    ConstrainDouble    zoom;
+    ConstrainDouble              focusDistance;
+    (boolean or ConstrainDouble) pan;
+    (boolean or ConstrainDouble) tilt;
+    (boolean or ConstrainDouble) zoom;
 
-    ConstrainBoolean   torch;
+    ConstrainBoolean             torch;
   };
 </pre>
 
@@ -581,7 +581,9 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>See <a>contrast</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>pan</code></dfn></dt>
-  <dd>See <a>pan</a> constrainable property.</dd>
+  <dd>A value of true is equivalent to a value of empty {{ConstrainDouble}}.
+      A value of false is equivalent to a value of null.
+      See <a>pan</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>saturation</code></dfn></dt>
   <dd>See <a>saturation</a> constrainable property.</dd>
@@ -593,10 +595,14 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>See <a>focus distance</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>tilt</code></dfn></dt>
-  <dd>See <a>tilt</a> constrainable property.</dd>
+  <dd>A value of true is equivalent to a value of empty {{ConstrainDouble}}.
+      A value of false is equivalent to a value of null.
+      See <a>tilt</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>zoom</code></dfn></dt>
-  <dd>See <a>zoom</a> constrainable property.</dd>
+  <dd>A value of true is equivalent to a value of empty {{ConstrainDouble}}.
+      A value of false is equivalent to a value of null.
+      See <a>zoom</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>torch</code></dfn></dt>
   <dd>See <a>torch</a> constrainable property.</dd>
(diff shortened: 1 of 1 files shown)


riju:
@eehakkin : Let's add the JS example in [Examples Section](https://w3c.github.io/mediacapture-image/#examples)

riju:
@beaufortfrancois , @reillyeon, @kenchris  PTAL

index.bs
```diff
@@ -767,15 +773,21 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
 
-    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera.  The setting represents pan in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.</li>
+    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera.  The setting represents pan in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.
+
+    Any algorithm which uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/pan}} attribute which is not equivalent to null MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to "camera" and its panTiltZoom member set to true, and, optionally, consider its deviceId member set to any appropriate device's deviceId.</li>
```
beaufortfrancois: Shall we link to the `panTiltZoom` member?
And/or add a note at least that links to https://github.com/w3c/permissions/pull/204/files?

index.bs
```diff
@@ -767,15 +773,21 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
 
     <li><dfn>Focus distance</dfn> is a numeric camera setting that controls the focus distance of the lens.  The setting usually represents distance in meters to the optimal focus distance.</li>
 
-    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera.  The setting represents pan in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.</li>
+    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera.  The setting represents pan in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.
```
beaufortfrancois: Nits: Remove double spaces.

index.bs
```diff
@@ -767,15 +773,21 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
 
     <li><dfn>Tilt</dfn> is a numeric camera setting that controls the tilt of the camera.  The setting represents tilt in arc seconds, which are 1/3600th of a degree.  Values are in the range from –180*3600 arc seconds to +180*3600 arc seconds.  Positive values tilt the camera upward when viewed from the front, and negative values tilt the camera downward as viewed from the front.
 
+    Any algorithm which uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/tilt}} attribute which is not equivalent to null MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to "camera" and its panTiltZoom member set to true, and, optionally, consider its deviceId member set to any appropriate device's deviceId.
```
beaufortfrancois: Nits: Remove double spaces.

index.bs
```diff
@@ -767,15 +773,21 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+
     <div class="note">
       There is no defined order when applying <a>pan</a> and <a>tilt</a>, the UA is allowed to apply them in any order.  In practice this should not matter since these values are absolute, so order will not affect the final position.  However, if applying pan and tilt is slow enough, the order in which they are applied may be visually noticeable.
     </div></li>
 
-    <li><dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens.  The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1.  The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).</li>
+    <li><dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens.  The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1.  The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).
```
beaufortfrancois: Nits: Remove double spaces.

index.bs
```diff
@@ -522,27 +522,27 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
...
+    ConstrainDouble              colorTemperature;
+    ConstrainDouble              iso;
+
+    ConstrainDouble              brightness;
+    ConstrainDouble              contrast;
+    ConstrainDouble              saturation;
+    ConstrainDouble              sharpness;
+
+    ConstrainDouble              focusDistance;
+    (boolean or ConstrainDouble) pan;
```
beaufortfrancois: Can you walk through the rationale of this `boolean or ConstrainDouble`?
Why introducing this pattern now?

eehakkin: My rationale is to keep similarity with audio/video. So while it is possible to say `navigator.mediaDevices.getUserMedia({video: {}})`, it is maybe clearer to say `navigator.mediaDevices.getUserMedia({video: true})`. Similarly while should be possible to say `navigator.mediaDevices.getUserMedia({video: {pan: {}, zoom: {}}})`, it might be clearer to say `navigator.mediaDevices.getUserMedia({video: {pan: true, zoom: true}})`.

beaufortfrancois: I agree with you about the simplicity. It looks much better.
I'm wondering why this pattern wasn't used before. Maybe there's something we're missing.

index.bs
eehakkin: Makes sense.

index.bs

index.bs

index.bs

index.bs
beaufortfrancois: @yellowdoge maybe would know

index.bs
```diff
@@ -863,22 +888,48 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
...
+          // Map tilt to a slider element.
+          const input = document.querySelector('input[name="tilt"][type="range"]');
+          input.min = capabilities.tilt.min;
+          input.max = capabilities.tilt.max;
+          input.step = capabilities.tilt.step;
+          input.value = settings.tilt;
+          input.oninput = function(event) {
+            track.applyConstraints({advanced: [{tilt: event.target.value}]});
+          };
+          input.parentElement.hidden = false;
```
beaufortfrancois: I think we can remove the `hidden` part as it doesn't add that much value in the context of an example in a spec.

beaufortfrancois: Note that it wasn't present in the `pan` example section

index.bs
yellowdoge: @beaufortfrancois no particular reason, IIRC this spec predates the major clean up and homogeneisation of the constraints about 3-4y ago...? So perhaps it was just due an overhaul. IMHO this looks better & we should create implementation bugs.

beaufortfrancois: Thank you!

index.bs
```diff
@@ -835,7 +847,7 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
   Slightly modified versions of these examples can be found in e.g. <a href="https://codepen.io/collection/nLeobQ/">this codepen collection</a>.
 </div>
 
-## Update camera zoom and {{takePhoto()}} ## {#example1}
+## Update camera pan, tilt and zoom and {{takePhoto()}} ## {#example1}
 
 <div class="note">
   The following example can also be found in e.g. <a href="https://codepen.io/miguelao/pen/BLPzKx/left?editors=1010">this codepen</a> with minimal modifications.
```
beaufortfrancois: I believe we can now link to an updated sample

jan-ivar:
@eehakkin Was the reason for the new `true` semantics to be able to acquire the camera without altering the current pan, tilt and zoom values? Otherwise why not just:
```js
navigator.mediaDevices.getUserMedia({video: {pan: 0, tilt: 0, zoom: 1}});
```

eehakkin:
> @eehakkin Was the reason for the new `true` semantics to be able to acquire the camera without altering the current pan, tilt and zoom values?

Yes, exactly that. So with the new `true` semantics it is possible to acquire a pan-tilt-zoom camera (instead of a static camera) during getUserMedia without altering the current pan, tilt and zoom values and then later pan, tilt or zoom the acquired camera using applyConstraints without a new prompt (as there is no need to elevate the permission from a normal camera permission to a pan-tilt-zoom camera permission as the latter one was requested from the beginning).

PR #221, opened 2020-04-15
Add PTZ explainer

beaufortfrancois:


ptz-explainer.md
```diff
@@ -0,0 +1,166 @@
...
+- fbeaufort@google.com
+
+## Participate
+
+- https://github.com/w3c/mediacapture-image/issues
+
+## Introduction
+
+[Some cameras] have the ability to pan and tilt which is especially useful in
+video conferencing (e.g. steering the camera to face the speaker).
```
riju, line 14: nit: Remove the (steering the camera) as this is in the Goals section.

beaufortfrancois: Done.

riju:
@kenchris : Do you want to have a look ?
@reillyeon, @eehakkin and I discussed this offline.

ptz-explainer.md
```diff
@@ -0,0 +1,166 @@
...
+in the "Interaction with the Permissions API" section below. If not granted,
+user will be prompted. Otherwise, the pan and tilt values passed in
+`applyConstraints()` will be used to move the camera.
+
+The example below shows how camera pan and tilt could be requested and presented
+to the user.
+
+```js
+// User is prompted to grant camera PTZ access only if the camera supports PTZ.
+// If the camera does not support PTZ, it falls back to a "regular" camera prompt.
```
kenchris, line 60: But what if the user doesn't want anyone to remotely control their camera but are fine with accepting a video stream, but just not PTZ features? I think it would be nice to tread this as a "Yes, I accept to share my camera, but disallow people controlling it remotely" - the explainer could discuss this option and it would be great if the API would allow that as the same call. Just fallback to regular camera, and not regular camera prompt.

Like

"site would like to use your camera and additionally control it remotely [block] [allow] [allow just camera]"

Or something like

```
Site would like to use your camera.

[x] Also, allow site to remotely control your camera

[block] [allow]
```

ptz-explainer.md
```diff
@@ -0,0 +1,166 @@
...
+  name: "camera",
+  panTiltZoom: true,
+});
+
+if (panTiltZoomPermissionStatus.state == "granted") {
+  // User has granted access to this website to control camera PTZ.
+}
+```
+
+To be explicit, `{name: "camera", panTiltZoom: true}` is stronger than `{name:
```
kenchris, line 132: I don't see exactly how you would be able to group them like in my examples above. If that is posssible, it would be good to spell it out. It seems right now that I will get one failed prompt and then have to reprompt, which will just result in the prompt-fatigue argument

kenchris:
From the TAG discussions, I think I can say that we care more about how this is presented to the user and about user privacy more than how it is coded by the developer.

Multiple prompts (one right after the other) is bad as it just creates more prompt fatigue. Some feel strong about that.

I would prefer a "this is what the site would like " and then "this is what the site got" approach - one prompt only".

@riju ague that this depends on permission grouping which is a future thing, but the counter argument would be to not do this feature before that feature is ready then, which is surely not the outcome you are looking at. Anyway, I don't see this really as grouped permissions but more like subpermissions.

As I read the current explainer example: if the user has PTZ camera, the prompt will ask for camera + PTZ support and the user will need to reject both... then the site will need to reprompt for just camera

That is a re-prompting scenario and one that is hard/impossible to turn into a single prompt without waiting for some magical future feature with unknown release date or success ratio.

Maybe it is possible to return a "partly-granted" permission and then allow the devs to query what is actually available or similar - just thinking outside the box here. This needs some more thought.

(I assume  you could just return granted today and then query - `if ("pan" in capabilities)` - and maybe that is fine, or maybe it is confusing unless it is clear that those are subpermissions. That would be a good discussion section for the explainer though. We really like those - alternatives considered sections.

I think that it is safe to say that we in the TAG care about avoiding prompt fatigue and making sure the user is in control in relation to privacy. This seems to solve one but not the other.

riju:
@engedy : we would like to hear your opinion on this (avoiding prompt fatigue) - 
If the user has PTZ camera, the prompt will ask for camera + PTZ, but user denies - How should UA interpret presently ? There's no partial-grant that we can interpret user wants us to fallback to normal camera permission, maybe user actually did not want to grant permission to camera at all. Is the checkbox permission presently in pipeline in Chrome ?

For more context, Balasz (privacy/UX team's) initial thinking was-

- Pan-Tilt-Zoom will be listed in the permission prompt only if supported, and we expect that those users specifically purchased the camera because of this capability, and then want to take advantage of the feature.

- We expect that explicitly listing Pan-Tilt-Zoom in the permission prompt will sufficiently discourage trustworthy sites from asking for it without reason.
 
- We will monitor potential abuse through UKM.

Checkbox and Grouped Permissions are something we can look in the future and those are UA specific.

I think we can add a NOTE in the spec, but actual prompt (message) are discretion of the UA.

reillyeon:
@kenchris The explainer doesn't say that the user will be prompted again. In the example interface there is no checkbox to deny PTZ permission but if a UA implemented this then the site would observe this as identical to gaining access to a camera that does not support PTZ. No re-prompting necessary. This should be added as a note.

From a Chrome perspective the current UX does not provide a checkbox but this is something that can be added without changing the API based on user feedback or if we observe abusive behavior.

kenchris:
Great to hear. It would be very good to clarify that in the explainer and add a note to the spec text

beaufortfrancois:
@kenchris I've updated explainer at https://github.com/w3c/mediacapture-image/pull/221/commits/22c1e02786096a777ba085f25725be4ae78c331b

ptz-explainer.md
```diff
@@ -31,22 +31,25 @@ steering the camera to face a speaker for instance.
...
 live video feed. Those will also be used in `getUserMedia()` to express whether a
 website wants to control camera PTZ functionality. In other words, it will be
-used to request the PTZ permission, a separate camera permission, in a single
-call. If the selected/connected camera does not support PTZ though, the PTZ
-permission won’t be requested and may fallback to the camera permission. If the
-user denies the PTZ permission, it may also fallback to the camera permission.
+used to request the PTZ permission, a separate permission, in a single
+`getUserMedia()` call, along the camera permission. If the selected/connected
+camera does not support PTZ though or user blocks solely the PTZ permission, the
+UA may fall back to the camera permission.
```
kenchris, line 46: I don't think that is a good recommendation as people might plug and unplug cameras with such support. I do that as I share cameras between computers

ptz-explainer.md
```diff
@@ -110,17 +113,22 @@ Having a separate PTZ permission allows the UA to differentiate between normal
 camera permissions and PTZ camera permissions as PTZ needs to be explicitly
 requested as an extension to the camera permission.
 
-UA may group both the camera permission request and the PTZ permission request.
-See mock below.
+The UA may group both the camera permission request and the PTZ permission
+request. See mock below.
```
kenchris, line 117: In a single permission prompt, for instance allowing user to pick and choose sub-permissions like PTZ

ptz-explainer.md
```diff
@@ -110,17 +113,22 @@ Having a separate PTZ permission allows the UA to differentiate between normal
...
 
 ![Mock of a PTZ permission prompt](/images/ptz-prompt-mock.png)
 
+Users can block and revoke the PTZ permission solely in UA settings. The UA may
+also provide a way for users to block PTZ directly from the prompt. Web
+developers can monitor those PTZ permission changes with the [permissions API].
```
kenchris, line 123: Example code?

ptz-explainer.md
```diff
@@ -0,0 +1,184 @@
...
+
+The current MediaStream Image Capture API already defines the `zoom` media track
+constraint. This feature is about adding `pan` and `tilt` and finding a
+consistent way for users to deal with PTZ.
+
+## Stakeholder Feedback / Opposition
+
+N/A
+
+## References & acknowledgements
```
kenchris, line 170: Remove second e in acknowledgment (US spelling :))

ptz-explainer.md
beaufortfrancois: That's basically what I'm saying below right?
> The UA may also provide a way for users to block PTZ directly from the prompt.

ptz-explainer.md
beaufortfrancois: That's why I've set "may fall back" as it's up to the UA. If it turns out to be unsuccessful, we should update the explainer indeed.

andypaicu:
> @engedy : we would like to hear your opinion on this (avoiding prompt fatigue) -
> If the user has PTZ camera, the prompt will ask for camera + PTZ, but user denies - How should UA interpret presently ? There's no partial-grant that we can interpret user wants us to fallback to normal camera permission, maybe user actually did not want to grant permission to camera at all. Is the checkbox permission presently in pipeline in Chrome ?

Hello, Andy here from the Chrome permissions team.

I agree it is up to the UA to provide whatever degree of control to their user when prompting them to make the decision. So I think that the spec itself should perhaps make suggestions at most as to how this is presented to the user.

For Chrome specifically the current status quo is that both Camera and PTZ will be blocked when the user blocks which means that the site won't be able to re-prompt for only Camera afterwards (and it won't happen automatically either, I think there was some confusion around this point). This is similar to how Mic + Camera behaves now, if a user might be comfortable with only giving Microphone access to a conferencing site they are still able to do so via the surfaces that allow them to change the permission, after the prompt. But that has some issues with discoverability.

Overall I don't think this is ideal, but I'm relatively fine with it. A checkbox-type solution allowing users to control directly from the prompt would probably be best, but at the same time I don't know how high priority it is since the user still has the ability to do this from the page info, and we're doing some work that will hopefully improve the discoverability of said page info sheet.

beaufortfrancois:
As discussed offline with @kenchris, I'm gonna merge

PR #228, opened 2020-06-15
Define when gUM rejects with required PTZ constraints

beaufortfrancois:
As discussed with @guidou, we should make it clear how `getUserMedia()` can reject when PTZ constraints are required.

This currently means `getUserMedia({video: {pan: 1}})` for instance rejects with an `OverConstrainedError`. We should think about how to make sure we don't leak camera PTZ capabilities when user didn't grant access yet.

@reillyeon @eehakkin @riju @andypaicu

### DIFF
--- a/ptz-explainer.md
+++ b/ptz-explainer.md
@@ -106,6 +106,19 @@ if ("tilt" in capabilities) {
 }
 ```
 
+The example below shows how camera pan could be reset when acquiring a
+PTZ camera in `getUserMedia()`.
+
+```js
+// User is prompted to grant both camera and PTZ access in a single call.
+// If the camera does not support PTZ or user denies PTZ permission, it fails
+// as PTZ constraints are required.
+const videoStream = await navigator.mediaDevices.getUserMedia({
+  // [NEW] Website asks to reset camera pan.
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
Done in https://github.com/w3c/mediacapture-image/pull/228/commits/9ab66353a519cb6730ce9341aca6665023cc85a1

beaufortfrancois:
@reillyeon The issue I'm seeing is that we would currently reject the `getUserMedia()` promise with an `OverConstrainedError` if camera does not support PTZ while we should ask user to grant PTZ access first.

beaufortfrancois:
I guess this is an implementation issue actually.

@riju If that looks good to you as well, please merge.

riju:
Looks good, I guess we need to keep a track of this [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=865799), which can help Chrome in not leaking information.

riju:
Will track https://github.com/w3c/mediacapture-main/issues/697

beaufortfrancois:
And I've created https://github.com/w3c/mediacapture-image/issues/229

PR #231, opened 2020-06-30
Add fingerprinting section to PTZ explainer

beaufortfrancois:


ptz-explainer.md
```diff
@@ -171,6 +171,30 @@ hardware supports it). The permission will have to be re-requested through
...
+   mandatory constraints so that the immediate failure of a `getUserMedia` call
+   with `OverConstrainedError` returns information about camera devices on the
+   system without prompting the user. This increases the surface available for
+   fingerprinting as already raised in the [Media Capture and Streams
+   spec](https://www.w3.org/TR/mediacapture-streams/#privacy-and-security-considerations).
+   The browser could mitigate this issue by always treating pan, tilt, and zoom
+   constraints as "ideal" in `getUserMedia` as suggested in
+   [#229](https://github.com/w3c/mediacapture-image/issues/229).
+   
+1. A malicious website could set pan, tilt, and zoom to minimally different values
```
yoavweiss, line 194: I think this scenario can be expanded beyond just an evil.com maintaining its own state into an evil.com 3P plug-in that enables leaking the HW state across origins.

E.g., if origins A and B both simultaneously open media streams where A sets PTZ to values {X, Y, Z}, while B reads the current PTZ state, A can leak a large number of bits to B, enabling correlating user IDs between origins.

If these bits are only accessible *after* the top-level origin has permissions to use the camera, that may not be that bad. Still worthwhile mentioning IMO.

I'm also interested in @jensenpaul's opinions on this.

youennf: > if origins A and B both simultaneously open media streams where A sets PTZ to values {X, Y, Z}, while B reads the current PTZ state

It seems not great that two different web pages get access to the PTZ at the same time. Should that be prevented? Safari is for instance muting a page capture if a new page starts to capture.

beaufortfrancois: Thanks for the hint @youennf 
I've just tried it and Safari indeed "pauses" camera feed when camera is accessed from another page to avoid this issue. Note that this is not exposed to web developers though: the mediastreamtrack is not "ended".

beaufortfrancois: I'm about to add this to the explainer. WDYT?

`1. Websites from different origins could share pan, tilt, and zoom real-time values during a media session. A way of mitigating this would be to allow only one origin at a time to access and control camera PTZ.`

ptz-explainer.md
```diff
@@ -171,6 +171,30 @@ hardware supports it). The permission will have to be re-requested through
...
+## Fingerprinting
+
+User fingerprinting is the practice of gathering multiple bits of user
+information from multiple sources (built-in hardware, user settings, installed
+peripherals, browsing data) and intersecting them together to create a unique
+signature of the user, that would enable to recognize them later on, even if
+they clear state from their browsers.
+
+1. Pan, tilt, and zoom hardware capabilities (e.g. `min`, `max`, `step`) and
+   current settings are not exposed to websites unless the user explicitely
```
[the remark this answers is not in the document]
beaufortfrancois: Done.

ptz-explainer.md
youennf: > Note that this is not exposed to web developers though: the mediastreamtrack is not "ended".

This is exposed using mute/unmute events. User Agent UI allows unmuting a given page, hence why tracks do not get ended.

Your description sounds like a good addition to me. The notion of origin is a bit vague for third-party iframes in that context but that could be refined later on. I would also mention the possibility to restrict to one top-level browsing context at a time.

yoavweiss: Thanks @youennf! Pausing the input (from camera, PTZ, etc) makes sense to me as a proper mitigation. Does that mean that only the foreground tab gets the actual stream input?

In the case of PTZ, it would also mean that the values should be reset to the ones previously set by the origin that can access the data. Essentially, we should treat the PTZ settings as a cached value and isolate that cache to the top-level origin.

youennf: > Does that mean that only the foreground tab gets the actual stream input?

Safari does not automatically mute/unmute when user switches tabs but I could see other implementations doing so. This is User Agent land really. The spec could say when to mute (for instance if PTZ capture starts/restarts) and leave the exact unmute heuristic to the User Agent.

beaufortfrancois:
Note that I've started a separate PR for the security section of the PTZ explainer at https://github.com/w3c/mediacapture-image/pull/232

ptz-explainer.md
beaufortfrancois: @youennf Based on your feedback I've added https://github.com/w3c/mediacapture-image/pull/231/commits/e39be06ad25e2340098d57aa2dd4bfaae62e8187. WDYT?

beaufortfrancois:
@riju please merge.
It looks good to @reillyeon and @engedy as discussed offline.

One nit, we may want to rename "fingerprinting" section to "privacy mitigations" in the future.

Issue #236, opened 2020-07-18
Reject applyConstraints with SecurityError if page is not visible for PTZ


beaufortfrancois:
This issue tracks spec change to make sure `applyConstraints()` rejects with `SecurityError` if the page is not visible for `pan`, `tilt`, and `zoom` constraints only.
The PTZ explainer will have to be updated as well.

PR #237, opened 2020-07-21
Reject applyConstraints if page is not visible for PTZ

beaufortfrancois:
This PR updates spec and explainer to reject applyConstraints with SecurityError if page is not visible for pan, tilt, and zoom.

FIX https://github.com/w3c/mediacapture-image/issues/236

### DIFF
--- a/ptz-explainer.md
+++ b/ptz-explainer.md
@@ -177,6 +177,9 @@ Requesting the PTZ permission is gated by well-known anti-abuse mechanisms:
 - starting a media session is available only in secure contexts,
 - the user has to explicitly allow PTZ through a permission prompt.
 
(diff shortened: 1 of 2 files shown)


ptz-explainer.md
```diff
@@ -177,6 +177,9 @@ Requesting the PTZ permission is gated by well-known anti-abuse mechanisms:
 - starting a media session is available only in secure contexts,
 - the user has to explicitly allow PTZ through a permission prompt.
 
+The page must be visible when the website updates camera pan, tilt, and zoom
+with `applyConstraints()`.
```
riju, line 181: else it should fail with `SecurityError`.

beaufortfrancois: Fixed. Thanks!

PR #247, opened 2020-08-18
Fix arc second range presentations

eehakkin:
Pan and tilt ranges should be from -180\*3600 to 180\*3600 arc seconds instead of from -180*3600* to 180*3600* arc seconds.


beaufortfrancois:
LGTM. Thanks for catching!

beaufortfrancois:
@riju Can you apply PTZ label and merge?

riju (approved):
:+1:

Issue #245, opened 2020-08-14
Sanitize PTZ capabilities/settings based on PTZ permission


beaufortfrancois:
As raised by @youennf in https://github.com/w3c/mediacapture-image/issues/243#issuecomment-674027572, the spec should say clearly that PTZ capabilities are not available when PTZ permission is not granted. Same for PTZ settings.

@eehakkin Can you take care of this?

youennf:
To be clear, it might be ok to allow web page to know that a selected camera has PTZ so that web page asks more permission to user. This is up to discussion but the spec should clearly specify it either way.

For PTZ settings, spec should clearly mandate sanitization.

beaufortfrancois:
For info, Chromium implementation sanitizes PTZ capabilities and settings based on PTZ permission.

youennf:
The particular case of PTZ capabilities seems worth getting input from the WebRTC WG.

eehakkin:
So the corresponding pull request is https://github.com/w3c/mediacapture-image/pull/248.

eehakkin:
> As raised by @youennf in [#243 (comment)](https://github.com/w3c/mediacapture-image/issues/243#issuecomment-674027572), the spec should say clearly that PTZ capabilities are not available when PTZ permission is not granted. Same for PTZ settings.

The both changes have now been merged.

beaufortfrancois:
Thanks! I'll close it.
@youennf Feel free to re-open if there's something we've missed.

jan-ivar:
This issue has a resolution associated with it in the [minutes](https://www.w3.org/2020/09/15-webrtc-minutes.html#r06) so I think it should remain open until its PR has been merged.

PR #248, opened 2020-08-18
Clarify pan/tilt/zoom capabilities and settings (#245)

eehakkin:
This CL mandates PTZ capabilities and settings must be sanitized based on permission.
I hope that this clarifies the issue.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,7 +493,10 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
-  <dd>This reflects the <a>pan</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>pan</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
+  In that case the UA MUST NOT expose the <a>pan</a> value range.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>saturation</code></dfn></dt>
   <dd>This reflects the permitted range of <a>saturation</a> setting. Values are numeric. Increasing values indicate increasing saturation.</dd>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -673,7 +682,10 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
   <dd>This reflects the current <a>contrast</a> setting of the camera.</dd>
 
   <dt><dfn dict-member for="MediaTrackSettings"><code>pan</code></dfn></dt>
-  <dd>This reflects the current <a>pan</a> setting of the camera.</dd>
+  <dd>This reflects the current <a>pan</a> setting of the camera.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
```
youennf, line 687: It seems checking the permission value at the time getSettings is called is sufficient.
Why adding a check for the permission at the time the track was created?

index.bs
```diff
@@ -493,7 +493,10 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
-  <dd>This reflects the <a>pan</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>pan</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
+  In that case the UA MUST NOT expose the <a>pan</a> value range.</dd>
```
youennf, line 499: I would split this PR in two, one for getSettings and one for getCapabilities.
This change would disallow some step-by-step permission flow (first request camera permission, then request PTZ permission if PTZ support)

index.bs
eehakkin: This PR does not add a check for the permission at the time the track was created. That check already exists in the spec. So reusing the result of that check (or the fact that the check was skipped because PTZ was not requested) seems sufficient.

Checking the permission value at the time getCapabilities or getSettings is called breaks the developer assumptions that availability of capabilities and settings do not change during track life time (like it is common to create UI components and their event listeners based on capabilities when the track starts).

I also think that PTZ capabilities and settings should not suddenly appear to non-PTZ tracks just because the origin gains PTZ permission through another getUserMedia call during the life time of the non-PTZ track.

What happens if the PTZ permission is revoked is more interesting. One option is that the track could be ended (in which case getCapabilities and getSettings do not have to check permissions). The second option is that the settings would not be updated after the revocation (in which case getCapabilities and getSettings again do not have to check permissions). The third option is that PTZ capabilities and settings would be removed (but would that require some kind of event so that the web page could listen for that event?).

index.bs
eehakkin: I will two separate PRs.

The idea (and way it is currently implemented in Chrome) is that capabilities are guarded by the PTZ permission so that PTZ ranges cannot be used for fingerprinting without explicit permission grant.

eehakkin:
Abandoned for https://github.com/w3c/mediacapture-image/pull/249 and https://github.com/w3c/mediacapture-image/pull/250.

index.bs
beaufortfrancois: Let's continue at https://github.com/w3c/mediacapture-image/pull/250

index.bs
beaufortfrancois: Let's continue this discussion at https://github.com/w3c/mediacapture-image/pull/249

Issue #229, opened 2020-06-18
Document OverConstrainedError for PTZ in explainer


beaufortfrancois:
As discussed at https://github.com/w3c/mediacapture-main/issues/697, the resolution to the fingerprinting/capabilities query in `getUserMedia()` in general seems to be to keep things as they are. It means that pan/tilt/zoom properties will be three more properties among the many that `getUserMedia()` already exposes.

In other words, `getUserMedia()` will reject with an `OverConstrainedError` if camera does not support PTZ and PTZ constraints are required without user prompt.

I believe we should document this in the upcoming privacy section of the PTZ explainer.

@riju @reillyeon @eehakkin @andypaicu @guidou

youennf:
I did not follow PTZ discussions, so I might lack some background here.
Adding more mandatory constraints is adding fingerprinting surface so I would hope we do not go there.

In general, I think we want to move away from mandatory constraints.
Can we define these new constraints so that they are only treated as ideal?

youennf:
This issue is probably worth some 'privacy' label

beaufortfrancois:
> Can we define these new constraints so that they are only treated as ideal?

@youennf  Do we have examples of constraints that are only treated as ideal?

youennf:
Adding the necessary hooks in media capture main is possible.
Before we start discussing it there, I would like to validate that restricting this constraint to 'ideal' is good enough for PTZ cameras. Is that correct?

youennf:
At least ideal constraints for getUserMedia.

beaufortfrancois:
Restricting pan, tilt, and zoom constraints in `getUserMedia` to be "ideal" would work.
In this case, the PTZ capability request would be optional, not mandatory.

```js
// [NEW] Does not fail with OverConstrainedError if no PTZ camera. 
await navigator.mediaDevices.getUserMedia({ video: { pan: true } });

// [NEW] Both would be equivalent. 
await navigator.mediaDevices.getUserMedia({ video: { pan: true } });
await navigator.mediaDevices.getUserMedia({ video: { advanced: [{ pan: true }] } });
```

It seems like a bigger change than PTZ according to https://github.com/w3c/mediacapture-main/issues/697

beaufortfrancois:
For info, https://github.com/w3c/mediacapture-image/pull/231 documents the current OverConstrainedError in the PTZ explainer.

beaufortfrancois:
@youennf What would be the next steps?

[...]

PR #250, opened 2020-08-19
Clarify pan/tilt/zoom capabilities (#245)

eehakkin:
Split from https://github.com/w3c/mediacapture-image/pull/248 per @youennf's request.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,7 +493,15 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
-  <dd>This reflects the <a>pan</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>pan</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
+  In that case the UA MUST NOT expose the <a>pan</a> value range.
+
+  <div class="note">
+    It does not matter whether the algorithm to <a>request permission to use</a> a PermissionDescriptor returns the current permission state without user interaction (when the current state is not "prompt") or asks the user's permission.
+    In both cases, the permission to use is requested and either granted or denied.
+  </div></dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>saturation</code></dfn></dt>
   <dd>This reflects the permitted range of <a>saturation</a> setting. Values are numeric. Increasing values indicate increasing saturation.</dd>
@@ -505,10 +513,16 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>This reflects the <a>focus distance</a> value range supported by the UA.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>tilt</code></dfn></dt>
-  <dd>This reflects the <a>tilt</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>tilt</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>tilt</a>.
+  In that case the UA MUST NOT expose the <a>tilt</a> value range.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>zoom</code></dfn></dt>
-  <dd>This reflects the <a>zoom</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>zoom</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>zoom</a>.
+  In that case the UA MUST NOT expose the <a>zoom</a> value range.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>torch</code></dfn></dt>
   <dd>A boolean indicating whether camera supports <a>torch</a>  mode- on meaning supported.</dd>


index.bs
```diff
@@ -505,10 +508,16 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
...
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>tilt</code></dfn></dt>
-  <dd>This reflects the <a>tilt</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>tilt</a> value range supported by the UA and by the track.
+
+  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>tilt</a>.
+  In that case the UA MUST NOT expose the <a>tilt</a> value range.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>zoom</code></dfn></dt>
-  <dd>This reflects the <a>zoom</a> value range supported by the UA.</dd>
+  <dd>This reflects the <a>zoom</a> value range supported by the UA by the track.
```
beaufortfrancois, line 517:
```suggestion
  <dd>This reflects the <a>zoom</a> value range supported by the UA and by the track.
```

beaufortfrancois (commented):
@youennf 

This PR reflects the current state of the [explainer](https://github.com/w3c/mediacapture-image/blob/master/ptz-explainer.md#control-camera-pantilt)

> If the PTZ permission is not granted, the pan, tilt, and zoom are not available as capabilities, settings, and constraints, even if the camera supports PTZ.

It is also reiterated in the [fingerprinting section](https://github.com/w3c/mediacapture-image/blob/master/ptz-explainer.md#fingerprinting) of the explainer.

> 1. Pan, tilt, and zoom hardware capabilities (e.g. `min`, `max`, `step`) and
   current settings are not exposed to websites unless the user explicitly
   grants PTZ permission....

beaufortfrancois (commented):
@eehakkin IIRC, it looks like `getUserMedia({video: true})` for a site that was already been granted PTZ access won't expose PTZ capabilities. Is this intended?
I thought we would use the PTZ permission regardless of the originated PTZ permission request result.

beaufortfrancois:
@youennf Exposing PTZ camera capabilities regardless of the PTZ permission status would increase fingerprinting. If that's something you want to allow a step by step permission flow, we could make it optional in the spec.

```diff
- In that case the UA MUST NOT expose the <a>tilt</a> value range.
+ In that case the UA MAY NOT expose the <a>tilt</a> value range.
```

youennf:
> @youennf Exposing PTZ camera capabilities regardless of the PTZ permission status would increase fingerprinting.

We should look at whether it is an issue with regards to passive and active fingerprinting.

We do not want to increase passive fingerprinting. Since PTZ setups are not widespread, it would be terrible to expose this capability without some kind of permission. MediaStreamTrack.getCapabilities is not possible since there is no capture track in that case. InputDeviceInfo.getCapabilities() should return an empty dictionary as per spec, so PTZ will not leak whatever the PTZ permission.

As of active fingerprinting (i.e. page is using the PTZ camera as a regular camera), we have some precedents to expose PTZ capabilities:
- Camera max resolution and frame rate are exposed
- In case a headset microphone is in use, the corresponding audiooutput device is exposed to the web without additional permission
- Camera device label is exposed and potentially leaks the exact device model

beaufortfrancois:
If that looks okay to other parties, I'll update explainer fingerprinting section and the example that currently relies on `getCapabilities()` instead of `getSettings()` to know whether PTZ can be controlled.

@jan-ivar @reillyeon @riju  What do you think?

beaufortfrancois:
Chrome privacy team got back to me and told me Chrome would expose nothing about the users camera having PTZ until the user has accepted a permission prompt that mentions PTZ.
In that case, I'd be in favor of updating this spec PR with "MAY" instead of "MUST" regarding exposing PTZ capabilities.  And I'll update the explainer.

@youennf How does that sound?

youennf:
Let's go with MUST for now. It is better to start strict and loosen things if there is enough benefit/interest.

This PR looks fine to me for MediaStreamTrack.getCapabilities.

As of InputDeviceInfo.getCapabilities, it would never expose PTZ even if PTZ permission is granted.
Maybe that is fine for now, given https://github.com/w3c/mediacapture-main/issues/672

eehakkin:
> @eehakkin IIRC, it looks like `getUserMedia({video: true})` for a site that was already been granted PTZ access won't expose PTZ capabilities. Is this intended?
> I thought we would use the PTZ permission regardless of the originated PTZ permission request result.

That is not intended.
I referred to the "request permission to use" algorithm which returns the current permission state if the state is not "prompt" (i.e. if the state is either "granted" or "denied"). So AFAIU if a site with PTZ access does `getUserMedia({video: true})`, that results in requesting permission to use a PermissionDescriptor with its name member set to `"camera"` and its panTiltZoom member set to `true` and the algorithm immediate returns `"grant"` and therefore PTZ capabilities will be exposed.

Do you have a better wording suggestions?

index.bs

beaufortfrancois:
Looking at it, I've misread it. We may want to add a note to clarify it for people like me ;)

beaufortfrancois (approved):

riju (approved):
Please add a note for better clarification.

eehakkin:
> Please add a note for better clarification.

Done.

youennf:
> So AFAIU if a site with PTZ access does `getUserMedia({video: true})`, that results in requesting permission to use a PermissionDescriptor with its name member set to `"camera"` and its panTiltZoom member set to `true` and the algorithm immediate returns `"grant"` and therefore PTZ capabilities will be exposed.

That is not my understanding of how things are specified.
If a page does getUserMedia({video: true}), UA will request permission to use with a descriptor { name: 'camera' }, but without any panTiltZoom member value. Then, getCapabilities will not be exposed according the current wording.
If you want that behavior, I would suggest to downscope the requirement to a MAY.

jan-ivar:
The current MUST wording seems most web compatible to me. Apps that rely on previously granted permission might work in Chrome, but not other browsers.

youennf:
Agreed MUST will forbid compat issues.
MAY is useful if some browsers would be willing to disclose PTZ capabilities if the website has camera permission.

jan-ivar:
Disclosing PTZ if the website only has camera permission would seem to undermine the permission model developed here and in the [ptz-explainer](https://github.com/w3c/mediacapture-image/blob/master/ptz-explainer.md), where _"Users can block and revoke the PTZ permission in UA settings"_.

These constraints may reveal the user to be in another web conference. Mitigations we've added for `width` and `height` in Firefox seem harder to repeat for `pan`, `tilt` and `zoom` e.g. if they control motors.

It seems simpler to demand sites specify e.g. `{pan: true}` every time for this info.

youennf:
> Disclosing PTZ

Are you referring to disclosing PTZ capabilities (i.e. the page knows a camera has PTZ support) or PTZ settings (i.e. the page can read/write actual device PTZ values)?
I am only referring to PTZ capabilities, not PTZ settings.

jan-ivar:
You're right I was thinking about settings, but I think it would be weird to expose one and not the other.

And since this spec doesn't appear to standardize the allowable ranges for `pan`, `tilt` and `zoom`, don't the capabilities leak device info as well?

youennf:
> And since this spec doesn't appear to standardize the allowable ranges for `pan`, `tilt` and `zoom`, don't the capabilities leak device info as well?

It could be implemented so that the only device info leak would be whether this is a PTZ camera or a regular camera:
If getUserMedia selects a regular camera, capabilities/settings would not list PTZ.
If getUserMedia selects a PTZ camera while PTZ is not requested, PTZ capabilities range would be [0,0] and PTZ setting values would be 0.
If getUserMedia selects a PTZ camera while PTZ is requested, PTZ capabilities and settings would be the actual values.

jan-ivar:
I think we'd want to specify anything like that. E.g. I might have interpreted `[0,0]` to mean fixed camera.

Is the use case a site that already has the camera and wants to know if it should show an "Enable ptz controls" button?

youennf:
> Is the use case a site that already has the camera and wants to know if it should show an "Enable ptz controls" button?

Yes, that is what I am thinking of.

youennf:
As discussed in https://github.com/w3c/mediacapture-image/issues/229#issuecomment-680931316, maybe we should have a panTiltZoom boolean constraint for getUserMedia. This could allow detection of PTZ cameras.
And then we would have pan/tilt/zoom specifically for applyConstraints, ignored by getUserMedia.

Issue #252, opened 2020-08-24
Stop referring to `advanced` constraints


jan-ivar:
This spec refers to the overly complex and outdated [advanced](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constraints-advanced) constraints syntax in:
1. [examples](https://w3c.github.io/mediacapture-image/#examples)
2. In issues like https://github.com/w3c/mediacapture-image/issues/243#issuecomment-674090590, where apparently one browser infers new meaning from their presence

The former is unnecessary, and the latter is a non-standard.

The differences between the older `advanced` constraints and the modern `ideal` constraints are [minimal](https://w3c.github.io/mediacapture-main/getusermedia.html#example-3) and best forgotten. TL;DR: they allowed for non-linear fallbacks between `width` x `height` x `frameRate` combos when `resizeMode` is `none`. They're largely redundant elsewhere, since most other constrainable capabilities are orthogonal to each other.

It's therefore best to not mention them, to discourage use. Maybe one day we'll be able to remove them from the platform.

beaufortfrancois:
How would we differentiate an optional PTZ request without advanced constraints?
See explainer.

(Sorry I'm on mobile and ooo this week)

jan-ivar:
@beaufortfrancois show me where in the spec `advanced` constraints let you do that.

jan-ivar:
@beaufortfrancois note constraints are non-required by default, so this should suffice:
```js
{video: {pan: true, tilt: true, zoom: true}}
```

jan-ivar:
Here's a good example of why `advanced` constraints suck: a likely bug in [ptz-explainer.md](https://github.com/w3c/mediacapture-image/blob/master/ptz-explainer.md#control-camera-pantilt):
```js
const videoStream = await navigator.mediaDevices.getUserMedia({
  video: {
    advanced: [{
      // [NEW] Website asks to control camera PTZ as well.
      pan: true, tilt: true, zoom: true,
    }],
  },
});
```
The above will, if the [constraints algorithm](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-selectsettings) is implemented correctly, only discover cameras that have _**all three capabilities**_.
If none exists, it'll return any regular camera (possibly forgoing cameras that have _some_ of these abilities but not all).

To get cameras with one or more of the `pan`, `tilt` and `zoom` abilities would instead have to be written like this:
```js
  video: {
    advanced: [
      {pan: true},
[...]

PR #253, opened 2020-08-24
Update examples using modern JS.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-image/issues/252 and users `async`/`await` throughout.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -901,76 +901,56 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
       &lt;label for="zoom">Zoom&lt;/label>
     &lt;/div>
     &lt;script>
-      var imageCapture;
+      let imageCapture;
 
-      navigator.mediaDevices.getUserMedia({
-          video: {pan: true, tilt: true, zoom: true}
-        })
-        .then(gotMedia)
-        .catch(err => console.error('getUserMedia() failed: ', err));
+      async function getMedia() {
+        try {
+          const stream = await navigator.mediaDevices.getUserMedia({
+            video: {pan: true, tilt: true, zoom: true},
+          });
+          const video = document.querySelector('video');
+          video.srcObject = stream;
 
-      function gotMedia(mediastream) {
-        const video = document.querySelector('video');
-        video.srcObject = mediastream;
+          const [track] = stream.getVideoTracks();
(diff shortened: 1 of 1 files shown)


jan-ivar:
[split ignore-whitespace diff](https://github.com/w3c/mediacapture-image/pull/253/files?diff=split&w=1)

index.bs
```diff
@@ -989,36 +969,34 @@ A {{Point2D}} represents a location in a two dimensional space. The origin of co
...
+          const canvas = document.querySelector('canvas');
+          const video = document.querySelector('video');
+
+          const stream = await navigator.mediaDevices.getUserMedia({video: true});
+          video.srcObject = stream;
+          const [track] = stream.getVideoTracks();
+          try {
+            const imageCapture = new ImageCapture(track);
+
+            stopButton.onclick = () => track.stop();
```
eehakkin, line 985: Shouldn't this stopButton variable be defined before use?

eehakkin (changes requested):
LGTM with nit.

index.bs
beaufortfrancois: `id="stopButton"` is enough in this case.

eehakkin: Oh, I did not know that.

eehakkin (approved):
LGTM.

riju (approved):
Thanks @jan-ivar for modernizing the examples. I see that the `advanced` constraints were removed rightly as this was not needed in this case. We can check in https://github.com/w3c/mediacapture-image/issues/252 for that.

beaufortfrancois:
I would not remove the `advanced` constraints in the applyConstraints examples as Chrome the only browser implementing it for now requires advanced constraints for using imageCapture constraints.
See https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/mediastream/media_stream_track.cc;l=720;drc=ce68dbf2ceede2e29df8ec69eba5fadd870704f4;bpv=1;bpt=1?q=applyConstraints%20f:media&ss=chromium

jan-ivar:
@beaufortfrancois Thanks for pointing that out. I've filed [crbug 1126045](https://bugs.chromium.org/p/chromium/issues/detail?id=1126045) on Chrome to update this to follow the spec.

Issue #256, opened 2020-08-31
Clarify meaning of PTZ constraints presence


jan-ivar:
This spec [appears to assume](https://w3c.github.io/mediacapture-image/#pan) any mention of `pan`, `tilt` or `zoom` will result in a PTZ device (& thus require ptz permission).

But that's not how the constraints algorithm works. E.g.
```js
navigator.mediaDevices.getUserMedia({video: {height: {min: 1080}, pan: true, tilt: true, zoom: true}});
```
...is equivalent to 
```js
navigator.mediaDevices.getUserMedia({video: {height: {min: 1080}, pan: {}, tilt: {}, zoom: {}}});
```
...where `{}` has no impact on [fitness distance](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-fitness-distance), and may choose your 1080p camera over your low-res PTZ camera.  

This seems unexpected, because a constraint value (even `true`) would usually carry weight, except as defined here.

You may also have granted the site more permission than necessary to access the device chosen.

### What's the model?

A related question is whether regular non-ptz cameras satisfy the following constraints:
```js
navigator.mediaDevices.getUserMedia({video: {pan: 0, tilt: 0, zoom: 0}});
```
Is it yes, because non-ptz cameras point front and center without zoom or tilt?
Is it no, because non-ptz cameras shouldn't have read-only `pan`, `tilt` or `zoom` settings? The spec doesn't say.

It appears we've overloaded the presence or absence of `pan`, `tilt` or `zoom` constraints with meaning.

### Options I see:
 1. Remove the `true` overload of `pan`, `tilt` and `zoom` constraints, and replace it with a `ptz` [ConstrainBoolean](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constrainboolean).
 2. Hotfix the fitness distance algorithm to give weight to presence of constraints.
 3. Make `true` a first-class value of `pan`, `tilt` and `zoom`, and specify a new `ConstrainULongOrBoolean`.

jan-ivar:
For context, this came up in https://github.com/w3c/mediacapture-image/issues/246#issuecomment-683777533.

youennf:
> where `{}` has no impact on [fitness distance](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-fitness-distance), and may choose your 1080p camera over your low-res PTZ camera.

However we define constraints and distance, the user agent can select any of the devices in finalSet.
[...]

Issue #246, opened 2020-08-18
Ignore PTZ request if no PTZ camera


beaufortfrancois:
As raised in https://github.com/w3c/mediacapture-image/issues/243#issuecomment-674108057 by @youennf, the spec should explicitly say that the PTZ request will be ignored and it will be treated like a regular camera request if there are no cameras with PTZ capabilities on the user machine during getUserMedia.

> > In the second query, if there are no cameras with PTZ capabilities, the PTZ request will be ignored and it will be treated like a regular camera request.
> 
> It seems you say that, if there is no PTZ camera, the permission request will be { name: "camera", panTiltZoom: false }.
> Where is this algorithm defined?

jan-ivar:
This risk of a prompt should be enough to deter trackers from calling `getUserMedia`.

It's unclear to me though whether we need to do the same for `applyConstraints`. Required constraints and `OverconstrainedError` are still arguably useful there.

eehakkin:
What exactly should be ignored?

If a web site calls `navigator.mediaDevices.getUserMedia({video: {pan: {}}})` and there are no PTZ cameras, the permission request will be `{name: "camera", panTiltZoom: false}` (instead of `{name: "camera", panTiltZoom: true}`) and what should happen then? Should the `pan: {}` part be completely ignored (an empty ConstrainDoubleRange value implies no constraints per the spec)?
At the moment Chrome implements the permission ignorance part but throws an OvercontstrainedError which does not make sense.

What if a web site calls `navigator.mediaDevices.getUserMedia({video: {pan: {ideal: 0}}})` and there are no PTZ cameras? Should be `pan: {ideal: 0}` part be completely ignored or should the [fitness distance](https://w3c.github.io/mediacapture-main/#dfn-fitness-distance) be 0 because a pan constraint does not apply to non-PTZ cameras? This choice has no effect to the selected device but affects how this should be specified in the spec. Or should the [fitness distance](https://w3c.github.io/mediacapture-main/#dfn-fitness-distance) be positive infinity because there is no actual pan setting? This would cause all non-PTZ cameras to be ignored and an OverconstrainedError to be thrown.

[...]

PR #257, opened 2020-09-15
Make true a first-class value for pan, tilt and zoom constraints

eehakkin:
This is Option 3 in https://github.com/w3c/mediacapture-image/issues/256 to clarify meaning of PTZ constraints presence.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -591,27 +591,27 @@ where actual is true if the <a>settings dictionary</a>'s value for the constrain
 
 <pre class="idl">
   partial dictionary MediaTrackConstraintSet {
-    ConstrainDOMString           whiteBalanceMode;
-    ConstrainDOMString           exposureMode;
-    ConstrainDOMString           focusMode;
-    ConstrainPoint2D             pointsOfInterest;
+    ConstrainDOMString       whiteBalanceMode;
+    ConstrainDOMString       exposureMode;
+    ConstrainDOMString       focusMode;
+    ConstrainPoint2D         pointsOfInterest;
 
-    ConstrainDouble              exposureCompensation;
-    ConstrainDouble              exposureTime;
-    ConstrainDouble              colorTemperature;
-    ConstrainDouble              iso;
+    ConstrainDouble          exposureCompensation;
+    ConstrainDouble          exposureTime;
+    ConstrainDouble          colorTemperature;
+    ConstrainDouble          iso;
 
-    ConstrainDouble              brightness;
-    ConstrainDouble              contrast;
-    ConstrainDouble              saturation;
-    ConstrainDouble              sharpness;
+    ConstrainDouble          brightness;
+    ConstrainDouble          contrast;
+    ConstrainDouble          saturation;
+    ConstrainDouble          sharpness;
 
-    ConstrainDouble              focusDistance;
-    (boolean or ConstrainDouble) pan;
-    (boolean or ConstrainDouble) tilt;
-    (boolean or ConstrainDouble) zoom;
+    ConstrainDouble          focusDistance;
+    ConstrainDoubleOrBoolean pan;
+    ConstrainDoubleOrBoolean tilt;
+    ConstrainDoubleOrBoolean zoom;
 
-    ConstrainBoolean             torch;
+    ConstrainBoolean         torch;
   };
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -528,6 +528,59 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
...
+      the exact required support for this property.</dd>
+</dd>
+
+  <dt><dfn dict-member for="ConstrainDoubleRangeOrBooleanParameters"><code>ideal</code></dfn></dt>
+  <dd>The ideal (target) value for this property or
+      the ideal (target) support for this property.</dd>
+</dl>
+
+<div class="note">
+An exact required or ideal no support for a property can be satisfied also by a device which is capable to support the property if only that capability is not exposed.
```
jan-ivar, line 554: What does "no support" mean here? Do you mean "constraint"?

I see this word a couple of places in this PR and they all look out of place to me, unless I'm misunderstanding something.

index.bs
```diff
@@ -791,32 +838,25 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
-
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+    Any algorithm which uses a pan constraint other than an exact required or an ideal no support constraint MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId.
```
jan-ivar, line 841: Why are we changing this prose? I would think any use of pan, tilt or zoom still needs to trigger a permission prompt.

I suggest we leave the permission prose alone in this PR since it doesn't seem like it needs to change to fix this issue, and leave it for https://github.com/w3c/mediacapture-image/issues/246.

jan-ivar (commented):
Thanks! This looks like the right direction to me, for the algorithm part (I'd leave the permission part for a different PR).

Unfortunately, due to a conflict, I won't be at the virtual interim today, so someone else will have to advocate for it. Note that @youennf is presenting a counter proposal, so be sure to bring this up if you prefer it.

eehakkin:
@jan-ivar: The consensus at the virtual interim yesterday was to throw Overconstrainted on requried values for PTZ (https://www.w3.org/2020/09/15-webrtc-minutes.html). I'll thus close this PR.

jan-ivar:
@eehakkin Why was this PR closed? This PR fixes fitness distance of ideal values, not `OverconstrainedError` on required values.

I don't see anything in the minutes about not fixing the fitness distance problem this spec has, which is https://github.com/w3c/mediacapture-image/issues/256.

I think we need to reopen this.

eehakkin:
Neither `ptz` media stream constraint (`navigator.mediaDevices.getUserMedia({video: {...}, ptz: true}`) nor `ConstrainDoubleOrBoolean` got support in the meeting.

`ConstrainDoubleOrBoolean` would allow:

- `navigator.mediaDevices.getUserMedia({video: {pan: true}})` and
  `navigator.mediaDevices.getUserMedia({video: {pan: {ideal: true}}})`
  These would do the same as what `navigator.mediaDevices.getUserMedia({video: {pan: true}})` and `navigator.mediaDevices.getUserMedia({video: {pan: {}}})` should do i.e. try to select a pan capable camera.
- `navigator.mediaDevices.getUserMedia({video: {pan: false}})` and
  `navigator.mediaDevices.getUserMedia({video: {pan: {ideal: false}}})`
  These would propably do the same as what `navigator.mediaDevices.getUserMedia({video: {pan: false}})` and `navigator.mediaDevices.getUserMedia({video: {}})` should do i.e. not expose pan capability.
- `navigator.mediaDevices.getUserMedia({video: {pan: {exact: true}}})` and
  `navigator.mediaDevices.getUserMedia({video: {pan: {exact: false}}})`
  These would throw a TypeError because required PTZ constraints should now do that.
- `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: {ideal: true}}]}})` and
  `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: {ideal: false}}]}})`
  But ideal advanced constraints are not useful (that is not limited to PTZ constraints applies to all constraints).
- `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: true}]}})` and
  `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: {exact: true}}]}})`
  These would do the same as what `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: true}]}})` and `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: {}}]}})` should do i.e. try to select a pan capable camera.
- `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: false}]}})` and
  `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: {exact: false}}]}})`
  These would propably do the same as what `navigator.mediaDevices.getUserMedia({video: {advanced: [{pan: false}]}})` and `navigator.mediaDevices.getUserMedia({video: {}})` should do i.e. not expose pan capability.

So `ConstrainDoubleOrBoolean` does not add anything useful.

That of course does not mean that the effect of `{pan: true}` on fitness distance or device selection should not be specified.

youennf:
I don't think we should introduce ConstrainDoubleOrBoolean.
As of fitness distance changes, we should first decide how we plan to handle PTZ device selection in case PTZ permission is granted.

PR #259, opened 2020-09-17
Allow empty PTZ capabilities without PTZ permission

beaufortfrancois:
This PR updates the PTZ explainer with the resolution "06" at https://www.w3.org/2020/09/15-webrtc-minutes.html#r06
I believe @eehakkin will follow with spec changes.

@jan-ivar @youennf @riju

### DIFF
--- a/ptz-explainer.md
+++ b/ptz-explainer.md
@@ -52,7 +52,8 @@ The [new "true" semantics] for `pan`, `tilt`, and `zoom` makes it possible to
 acquire a PTZ camera in `getUserMedia()` without altering the current pan, tilt
 and zoom values. If the PTZ permission is not granted, the `pan`, `tilt`, and
 `zoom` are not available as capabilities, settings, and constraints, even if the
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
(gentle ping)

youennf (approved):
LGTM.
Minor nit about whether UA can decide to expose only pan with empty values if camera is PTZ (but only has pan).
The intent is probably to expose all of them or none of them. This way, 1 bit of information is leaked instead of 3.

beaufortfrancois:
> Minor nit about whether UA can decide to expose only pan with empty values if camera is PTZ (but only has pan).
> The intent is probably to expose all of them or none of them. This way, 1 bit of information is leaked instead of 3.

Fixed in https://github.com/w3c/mediacapture-image/pull/259/commits/325fdb81de022131b731e324f916e1106a7120e8

riju (approved):
:+1:

PR #260, opened 2020-09-17
Allow PTZ support to be always exposed in capabilities

eehakkin:
Fixes https://github.com/w3c/mediacapture-image/issues/245. This is to address https://www.w3.org/2020/09/15-webrtc-minutes.html#r06

***
<a href="https://pr-preview.s3.amazonaws.com/eehakkin/intel-w3c-mediacapture-image/pull/260.html" title="Last updated on Feb 15, 2021, 7:19 AM UTC (3006eb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-image/260/72ee760...eehakkin:3006eb1.html" title="Last updated on Feb 15, 2021, 7:19 AM UTC (3006eb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,10 +493,16 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
-  <dd>This reflects the <a>pan</a> value range supported by the UA and by the track.
+  <dd>
+    This reflects the <a>pan</a> value range supported by the UA and by the track.
 
-  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
-  In that case the UA MUST NOT expose the <a>pan</a> value range.</dd>
+    If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
+    In that case the UA MUST NOT expose the <a>pan</a> value range but MAY provide an empty {{MediaSettingsRange}} dictionary to indicate that the underlying video source supports <a>pan</a>.
+
+    <div class="note">
+      Even if the UA provides an empty {{MediaSettingsRange}} dictionary to indicate that the underlying video source supports <a>pan</a>, <a>tilt</a> or <a>zoom</a>, that support can be utilized only after a new {{getUserMedia()}} call which resolves with a {{MediaStream}} object which contains a video track which supports <a>pan</a>, <a>tilt</a> or <a>zoom</a>.
+    </div>
+  </dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>saturation</code></dfn></dt>
   <dd>This reflects the permitted range of <a>saturation</a> setting. Values are numeric. Increasing values indicate increasing saturation.</dd>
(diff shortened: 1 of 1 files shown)


eehakkin:
@beaufortfrancois @jan-ivar @riju @youennf: PTAL.

index.bs
```diff
@@ -493,15 +493,21 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
...
-    It does not matter whether the algorithm to <a>request permission to use</a> a PermissionDescriptor returns the current permission state without user interaction (when the current state is not "prompt") or asks the user's permission.
-    In both cases, the permission to use is requested and either granted or denied.
-  </div></dd>
+    <div class="note">
+      It does not matter whether the algorithm to <a>request permission to use</a> a PermissionDescriptor returns the current permission state without user interaction (when the current state is not "prompt") or asks the user's permission.
+      In both cases, the permission to use is requested and either granted or denied.
+    </div>
+
+    <div class="note">
+      Even if the UA provides an empty {{MediaSettingsRange}} dictionary to indicate that the underlying video source supports <a>pan</a>, that support can be utilized only after a new {{getUserMedia()}} call which resolves a {{MediaStream}} object which contains a video track which supports <a>pan</a>.
```
beaufortfrancois, line 508: Shall we have this note applied to `tilt` and `zoom` as well?

beaufortfrancois (approved):
LGTM with nits

index.bs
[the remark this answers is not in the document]

beaufortfrancois:
(gentle ping) @youennf @jan-ivar

index.bs
```diff
@@ -493,15 +493,21 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
...
+    In that case the UA MUST NOT expose the <a>pan</a> value range but MAY provide an empty {{MediaSettingsRange}} dictionary to indicate that the underlying video source supports <a>pan</a>.
 
-  <div class="note">
-    It does not matter whether the algorithm to <a>request permission to use</a> a PermissionDescriptor returns the current permission state without user interaction (when the current state is not "prompt") or asks the user's permission.
-    In both cases, the permission to use is requested and either granted or denied.
-  </div></dd>
+    <div class="note">
+      It does not matter whether the algorithm to <a>request permission to use</a> a PermissionDescriptor returns the current permission state without user interaction (when the current state is not "prompt") or asks the user's permission.
+      In both cases, the permission to use is requested and either granted or denied.
```
youennf, line 504: This sentence is not very clear to me.

youennf (approved):
LGTM.
In the future, I would be interested in understanding whether the "MAY be exposed" can be consolidated as browsers implement this feature.

index.bs
eehakkin: In both cases (when the algorithm returns the current permission state without user interaction or when the algoithm asks the user's permission), the permission to use is requested. Thus the first condition above stating that _the track has been created without requesting permission to use a PermissionDescriptor with its name member set to camera and its panTiltZoom member set to true_ is not fulfilled. 
In the both cases, the permission is either granted or denied. If it is granted, the second condition above stating that _that permission request is denied_ is not fulfilled, and if it is denied, the second condition above is fulfilled.

How would you change that note?

jan-ivar: @eehakkin Why is the note there? Is it to warn readers the [request permission to use](https://w3c.github.io/permissions/#request-permission-to-use) algorithm, despite its name, doesn't necessarily prompt the user every time?

If so, then lets remove the note, since that's intrinsic behavior of the permissions spec, and not unique to imageCapture. Other specs don't have such notes when invoking this algorithm.

index.bs
```diff
@@ -493,15 +493,21 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
   <dd>This reflects the supported range of <a>contrast</a>. Values are numeric.  Increasing values indicate increasing contrast.</dd>
 
   <dt><dfn dict-member for="MediaTrackCapabilities"><code>pan</code></dfn></dt>
-  <dd>This reflects the <a>pan</a> value range supported by the UA and by the track.
+  <dd>
+    This reflects the <a>pan</a> value range supported by the UA and by the track.
 
-  If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
-  In that case the UA MUST NOT expose the <a>pan</a> value range.
+    If the track has been created without <a>requesting permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true or if that permission request is denied, the track does not support <a>pan</a>.
```
jan-ivar, line 499: This sentence seems to suggest the relevant permissions are the ones granted at the **_time of creation_** of the track, which seems wrong.

If the site sees `pan: {}`, decides to requests more permission with `applyConstraints({pan: true})`, and the user approves, then it should see `pan: {min: -648000, max: +648000}` from that point on.

To do that requires that we look at the _current_ permissions, i.e. from a fresh permission descriptor. Can you add that?

jan-ivar (changes requested):

index.bs
eehakkin: > This sentence seems to suggest the relevant permissions are the ones granted at the **_time of creation_** of the track, which seems wrong.

That's right, and it isn't wrong. Track capabilities are effectively constant and do not change during the life-time of the track.

> If the site sees `pan: {}`, decides to request more permission with `track.applyConstraints({pan: true})`, and the user approves, then it should see `pan: {min: -648000, max: +648000}` from that point on.

If the site sees `pan: {}`, it can decide to request more permission with `navigator.mediaDevices.getUserMedia({video: {deviceId: {exact: track.getSettings().deviceId}, pan: true})`, and if the user approves, then it should see `pan: {min: -648000, max: +648000}` from that point on.

If the site sees `pan: {}` (without min and max), the track has been created without `pan: true` (or tilt or zoom) thus the track does not support PTZ. Therefore, `track.applyConstraints({pan: true})` does nothing (because the ideal constraint cannot be fulfilled) and `track.applyConstraints({pan: {exact: 0}})` throws `OverconstrainedError` (because the required constraint cannot be fulfilled).

index.bs
eehakkin: > @eehakkin Why is the note there? Is it to warn readers the [request permission to use](https://w3c.github.io/permissions/#request-permission-to-use) algorithm, despite its name, doesn't necessarily prompt the user every time?

Yes, for that reason, exactly. It was requested at https://github.com/w3c/mediacapture-image/pull/250.

index.bs
youennf: > That's right, and it isn't wrong. Track capabilities are effectively constant and do not change during the life-time of the track.

This is the current consensus that was discussed during the interim, I believe for simplicity.
I am not opposed to revisiting it but this should be tackled as a separate issue.

eehakkin: If that is revisited, the main spec should probably also be changed because it specifies that capabilities are effectively constant.

index.bs
jan-ivar: @eehakkin where exactly was it requested (can you deep-link)?

I don't think this spec needs a note more than any other spec, but if we keep it I suggest clearer language, e.g.:

_"Note: The **request permission to use** algorithm, despite its name, does not necessarily prompt the user every time."_

index.bs
jan-ivar: I agree with @youennf here. @eehakkin mediacapture-main [says](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-capabilities): _"Source capabilities are effectively constant. Applications should be able to depend on a specific source having the same capabilities for any browsing session."_

I think we're fine because the underlying (source) capabilities aren't changing, we're just un-filtering them. There's [precedence](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-inputdeviceinfo-getcapabilities) for that.

The premise of this PR is to detect the availability of a PTZ device, so the site can act on it.

index.bs
eehakkin: Because this PR is not about that note, I created https://github.com/w3c/mediacapture-image/pull/266 for the removal.

PR #261, opened 2020-09-17
Reject with TypeError on required PTZ contraints in basic getUserMedia contraints

eehakkin:
This is to address https://www.w3.org/2020/09/15-webrtc-minutes.html#r05

Some details related to the resolution are not entirely clear to me, so please correct me if I misunderstood something or if you prefer something else.

## getUserMedia()

Required PTZ constraints in the basic constraint set are denied (and throw an `OverconstraintedError`) and PTZ capability constraints (`{zoom: true}` etc.) are ideal in the basic constraint set. Therefore, a PTZ constraint in the basic set cannot cause getUserMedia() to fail conditionally. This should be a clear case, I hope.

In this PR, required PTZ constraints are still allowed in the advanced constraint sets, for four reasons:

1. The advanced constraint sets cannot throw an OverconstraintedError but the UA may instead ignore them.
2. More importantly, a PTZ capability requests ({advanced: [{tilt: true}]} etc.) look like bare value constraints (while being syntactic sugar for empty constraints i.e. {advanced: [{tilt: {}}]} etc.) and bare value constraints are normally treated as exact required constraints in the advanced sets.
3. I do not see justification for added complexity required for changing PTZ constraint types to ConstraintDoubleOrBoolean which would allow {advanced: [{pan: {ideal: true}}]} etc. to be used.
4. The advanced constraint sets cannot be used for fingerprinting without a camera permission.

What do you think?

## applyConstraints()

Required PTZ constraints are still allowed. Web apps can check capabilities with getCapabilities() before applying constraints. While only ideal and exact required (but not min and max required) constraints probably make sense, I do not see a point for special-casing PTZ constraints here.

I also clarified that PTZ capabilities cannot be obtained with applyConstraints().

@beaufortfrancois @guidou @jan-ivar @riju @youennf: PTAL

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -596,9 +596,10 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>See <a>contrast</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>pan</code></dfn></dt>
-  <dd>A value of true is normalized to a value of empty {{ConstrainDouble}}.
+  <dd>A value of true is normalized to a value of empty {{ConstrainDoubleRange}}.
       A value of false is normalized to a value of undefined.
-      See <a>pan</a> constrainable property.</dd>
+      If the value after normalization is a double value or a {{ConstrainDoubleRange}} value (whether empty or not), that implies a <a>pan capability</a> request which is ideal in basic constraints and exact <a>required</a> in advanced constraints.
+      See <a>pan</a> and <a>pan capability</a> constrainable properties.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>saturation</code></dfn></dt>
   <dd>See <a>saturation</a> constrainable property.</dd>
@@ -610,14 +611,16 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for 
   <dd>See <a>focus distance</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>tilt</code></dfn></dt>
-  <dd>A value of true is normalized to a value of empty {{ConstrainDouble}}.
+  <dd>A value of true is normalized to a value of empty {{ConstrainDoubleRange}}.
       A value of false is normalized to a value of undefined.
-      See <a>tilt</a> constrainable property.</dd>
+      If the value after normalization is a double value or a {{ConstrainDoubleRange}} value (whether empty or not), that implies a <a>tilt capability</a> request which is ideal in basic constraints and exact <a>required</a> in advanced constraints.
+      See <a>tilt</a> and <a>tilt capability</a> constrainable properties.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>zoom</code></dfn></dt>
-  <dd>A value of true is normalized to a value of empty {{ConstrainDouble}}.
+  <dd>A value of true is normalized to a value of empty {{ConstrainDoubleRange}}.
       A value of false is normalized to a value of undefined.
-      See <a>zoom</a> constrainable property.</dd>
+      If the value after normalization is a double value or a {{ConstrainDoubleRange}} value (whether empty or not), that implies a <a>zoom capability</a> request which is ideal in basic constraints and exact <a>required</a> in advanced constraints.
+      See <a>zoom</a> and <a>zoom capability</a> constrainable properties.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>torch</code></dfn></dt>
   <dd>See <a>torch</a> constrainable property.</dd>
@@ -798,35 +801,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
 
     <li><dfn>Focus distance</dfn> is a numeric camera setting that controls the focus distance of the lens.  The setting usually represents distance in meters to the optimal focus distance.</li>
 
-    <li><dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera. The setting represents pan in arc seconds, which are 1/3600th of a degree. Values are in the range from –180\*3600 arc seconds to +180\*3600 arc seconds. Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.
+    <li>
+      <dfn>Pan</dfn> is a numeric camera setting that controls the pan of the camera. The setting represents pan in arc seconds, which are 1/3600th of a degree. Values are in the range from -180\*3600 arc seconds to +180\*3600 arc seconds. Positive values pan the camera clockwise as viewed from above, and negative values pan the camera counter clockwise as viewed from above.
 
-    Any algorithm which uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/pan}} dictionary member which exists after a possible normalization MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId.
+      <dfn>Pan capability</dfn> is a boolean camera property representing camera's ability to pan which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/pan}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw a {{TypeError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
+        <li>The algorithm MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId unless there are no pan capable cameras.
+            If that permission is not requested or if the permission is denied, the pan setting and capability ranges MUST NOT be exposed.</li>
+      </ul>
 
-    <div class="note">
-      The {{MediaTrackConstraintSet/pan}}, {{MediaTrackConstraintSet/tilt}} and {{MediaTrackConstraintSet/zoom}} dictionary members exists after a possible normalization if the normalized value is a double value or a {{ConstrainDoubleRange}} value (whether empty or not).
-      A boolean value of true is normalized to an empty {{ConstrainDoubleRange}} value thus the dictionary member exists after a normalization.
-      A boolean value of false is normalized to no value thus the dictionary member does not exist after a normalization.
-      An empty {{ConstrainDoubleRange}} value implies no constraints but only a permission and capability request.
-    </div></li>
+      If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a normalization.
 
-    <li><dfn>Tilt</dfn> is a numeric camera setting that controls the tilt of the camera. The setting represents tilt in arc seconds, which are 1/3600th of a degree. Values are in the range from –180\*3600 arc seconds to +180\*3600 arc seconds. Positive values tilt the camera upward when viewed from the front, and negative values tilt the camera downward as viewed from the front.
+      <div class="note">
+        The {{MediaTrackConstraintSet/pan}}, {{MediaTrackConstraintSet/tilt}} and {{MediaTrackConstraintSet/zoom}} dictionary members exist after a normalization if the normalized value is a double value or a {{ConstrainDoubleRange}} value (whether empty or not).
+        A boolean value of true is normalized to an empty {{ConstrainDoubleRange}} value thus the dictionary member exists after a normalization.
+        A boolean value of false is normalized to no value thus the dictionary member does not exist after a normalization.
+        An empty {{ConstrainDoubleRange}} value implies no pan, tilt or zoom constraints but only a permission and capability request.
+      </div>
+    </li>
 
-    Any algorithm which uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/tilt}} dictionary member which exists after a possible normalization MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId.
+    <li>
+      <dfn>Tilt</dfn> is a numeric camera setting that controls the tilt of the camera. The setting represents tilt in arc seconds, which are 1/3600th of a degree. Values are in the range from -180\*3600 arc seconds to +180\*3600 arc seconds. Positive values tilt the camera upward when viewed from the front, and negative values tilt the camera downward as viewed from the front.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/tilt}} dictionary member exists after a possible normalization.
+      <dfn>Tilt capability</dfn> is a boolean camera property representing camera's ability to tilt which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    <div class="note">
-      There is no defined order when applying <a>pan</a> and <a>tilt</a>, the UA is allowed to apply them in any order. In practice this should not matter since these values are absolute, so order will not affect the final position. However, if applying pan and tilt is slow enough, the order in which they are applied may be visually noticeable.
-    </div></li>
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/tilt}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw a {{TypeError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
+        <li>The algorithm MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId unless there are no tilt capable cameras.
+            If that permission is not requested or if the permission is denied, the tilt setting and capability ranges MUST NOT be exposed.</li>
+      </ul>
 
-    <li><dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens. The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1. The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).
+      If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/tilt}} dictionary member exists after a normalization.
 
-    Any algorithm which uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/zoom}} dictionary member which exists after a possible normalization MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId.
+      <div class="note">
+        There is no defined order when applying <a>pan</a> and <a>tilt</a>, the UA is allowed to apply them in any order. In practice this should not matter since these values are absolute, so order will not affect the final position. However, if applying pan and tilt is slow enough, the order in which they are applied may be visually noticeable.
+      </div>
+    </li>
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/zoom}} dictionary member exists after a possible normalization.
+    <li>
+      <dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens. The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1. The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).
 
+      <dfn>Zoom capability</dfn> is a boolean camera property representing camera's ability to zoom which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
+
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/zoom}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw a {{TypeError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
+        <li>The algorithm MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId unless there are no zoom capable cameras.
+            If that permission is not requested or if the permission is denied, the zoom setting and capability ranges MUST NOT be exposed.</li>
+      </ul>
+
+      If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/zoom}} dictionary member exists after a normalization.
+    </li>
 
     <li><dfn>Fill light mode</dfn> describes the flash setting of the capture device (e.g. `auto`, `off`, `on`). <dfn>Torch</dfn> describes the setting of the source's fill light as  continuously connected, staying on as long as {{track}} is active.</li>
   </ol>
@@ -1182,8 +1214,11 @@ urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: dictionary; text:
 
 urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: dictionary; text: MediaTrackSettings; url: idl-def-MediaTrackSettings
 
+urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: interface; text: OverconstrainedError; url: overconstrainederror-object
 
 urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: dfn; text: constrainable property; url: defining-a-new-constrainable-property
+
+urlPrefix: https://www.w3.org/TR/mediacapture-streams/#; type: dfn; text: required; url: dfn-required
 
 urlPrefix: https://www.w3.org/TR/page-visibility/; type: attribute; text: visibilityState; for: Document; url: dom-visibilitystate
 </pre>


index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+      <dfn>Pan capability</dfn> is a boolean camera property representing camera's ability to pan which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+      If the {{getUserMedia()}} algorithm uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/pan}} dictionary member which exists after a possible normalization:
+      <ul>
+        <li>The algorithm MUST throw an {{OverconstrainedError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
```
youennf, line 803: This is fine to state it here.
Note though that mediacapture-main will also state this as part of https://github.com/w3c/mediacapture-main/pull/707

index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
-      A boolean value of false is normalized to no value thus the dictionary member does not exist after a normalization.
-      An empty {{ConstrainDoubleRange}} value implies no constraints but only a permission and capability request.
-    </div></li>
+      If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
 
-    <li><dfn>Tilt</dfn> is a numeric camera setting that controls the tilt of the camera. The setting represents tilt in arc seconds, which are 1/3600th of a degree. Values are in the range from –180\*3600 arc seconds to +180\*3600 arc seconds. Positive values tilt the camera upward when viewed from the front, and negative values tilt the camera downward as viewed from the front.
+      <div class="note">
+        The {{MediaTrackConstraintSet/pan}}, {{MediaTrackConstraintSet/tilt}} and {{MediaTrackConstraintSet/zoom}} dictionary members exists after a possible normalization if the normalized value is a double value or a {{ConstrainDoubleRange}} value (whether empty or not).
```
reillyeon, line 811: Do you mean "exist" rather than "exists"?

index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+            If that permission is not requested or if the permission is denied, the pan setting and capability ranges MUST NOT be exposed.</li>
+      </ul>
 
-    <div class="note">
-      The {{MediaTrackConstraintSet/pan}}, {{MediaTrackConstraintSet/tilt}} and {{MediaTrackConstraintSet/zoom}} dictionary members exists after a possible normalization if the normalized value is a double value or a {{ConstrainDoubleRange}} value (whether empty or not).
-      A boolean value of true is normalized to an empty {{ConstrainDoubleRange}} value thus the dictionary member exists after a normalization.
-      A boolean value of false is normalized to no value thus the dictionary member does not exist after a normalization.
-      An empty {{ConstrainDoubleRange}} value implies no constraints but only a permission and capability request.
-    </div></li>
+      If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
```
reillyeon, line 808: "if its {{MediaTrackConstraintSet/pan}} dictionary member" (to match above)

index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/tilt}} dictionary member exists after a possible normalization.
+      <dfn>Tilt capability</dfn> is a boolean camera property representing camera's ability to tilt which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    <div class="note">
-      There is no defined order when applying <a>pan</a> and <a>tilt</a>, the UA is allowed to apply them in any order. In practice this should not matter since these values are absolute, so order will not affect the final position. However, if applying pan and tilt is slow enough, the order in which they are applied may be visually noticeable.
-    </div></li>
+      If the {{getUserMedia()}} algorithm uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/tilt}} dictionary member which exists after a possible normalization:
```
reillyeon, line 824: Same comment as for "pan" above.

index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+      <dfn>Pan capability</dfn> is a boolean camera property representing camera's ability to pan which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+      If the {{getUserMedia()}} algorithm uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/pan}} dictionary member which exists after a possible normalization:
```
reillyeon, line 801: "dictionary member exists after normalization"?

The word "which" seems unnecessary and I don't understand the significance of the word "possible".

index.bs
```diff
@@ -789,35 +792,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+    </li>
+
+    <li>
+      <dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens. The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1. The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/zoom}} dictionary member exists after a possible normalization.
+      <dfn>Zoom capability</dfn> is a boolean camera property representing camera's ability to zoom which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
+      If the {{getUserMedia()}} algorithm uses a {{MediaTrackConstraintSet}} object and its {{MediaTrackConstraintSet/zoom}} dictionary member which exists after a possible normalization:
```
reillyeon, line 844: Same comment as for "pan" above.

reillyeon (changes requested):

index.bs
eehakkin: So there are actually three conditions:

1. the `getUserMedia()` algorithm uses a `MediaTrackConstraintSet` object
2. the `getUserMedia()` algorithm uses the `pan` dictionary member of the `MediaTrackConstraintSet` object
3. the dictionary member exists after normalization

Dropping the word "which" would drop the second condition. I spelled out all three condition, instead. I hope it is more clear now.

I previously used the word "possible" because normalization is done only for boolean values but not for non-existing values, for double values nor for ConstrainDoubleRange values. It is not significant so I dropped it.

index.bs
eehakkin: Ack.

index.bs
eehakkin: I spelled out all three condition, instead.

index.bs
eehakkin: I spelled out all three condition, instead (to match above).

index.bs
eehakkin: Yes (the note was probably originally only for pan, hence exists). Fixed.

index.bs
eehakkin: Fixed.

index.bs
```diff
@@ -598,6 +598,7 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>pan</code></dfn></dt>
   <dd>A value of true is normalized to a value of empty {{ConstrainDouble}}.
       A value of false is normalized to a value of undefined.
+      An empty {{ConstrainDouble}} implies no constraints but a <a>pan capability</a> request (ideal in basic constraints and required in advanced constraints).
```
guidou, line 601: What does ideal in basic constraints and required in advanced constraints mean?
My understanding is that we don't want to have a way to require the capability.
Since advanced constraints are for required constraints only, pan:{} in advanced should only mean no constraints and nothing else. No capability request. Otherwise we do not solve the fingerprinting issue we want to address by not having required constraints.

To make it clear, allowing pan:{} or pan:true to mean a required capability request in an advanced set allows fingerprinting by adding additional constraints and checking if the other constraints were ignored.
For example: getUserMedia({video: {advanced: [{pan:true, width:10}]}).
In this case, in Chromium (which supports crop-and-scale), getting a track with a width different than 10 means that pan is not supported, which is basically equivalent to having an OverconstrainedError if the capability request was required in the basic set.

IMO, now that we want to treat capability requests similarly to an ideal value, such requests should be only in the basic set, which is the only one that allows ideal values.

eehakkin:
In the [intro](https://github.com/w3c/mediacapture-image/pull/261#issue-488934739), I stated that some details related to the resolution are not entirely clear to me and that this PR ensures that PTZ constraint cannot cause `getUserMedia()` to fail conditionally. In https://github.com/w3c/mediacapture-image/pull/261#discussion_r493634893 @guidou noted that that is not enough to prevent fingerprinting.

So I have more questions:

- What should `getUserMedia({video: {advanced: [{pan: {exact: 1234}, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: 1234, width: 10}]}` do? Should the advanced constraint set be ignored? Note that @youennf's PR https://github.com/w3c/mediacapture-main/pull/707 does not do that. If the advanced constraint set is ignored, should it be ignored silently, should it produce a console warning or equivalent or should it throw an `OverconstrainedError` (which advanced constraints do not normally throw)?
- What should `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` do? Is that ideal request for pan capability and for the pan setting of 1234?
- What should `getUserMedia({video: {advanced: [{pan: true, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: {}, width: 10}]}` do? Is that ideal request for pan capability (like `getUserMedia({video: {pan: true}})`)? Makes kind of sense. Or is that no constraint and no capability request? That would be easy but in that case is  `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` an ideal request for the pan setting of 1234 but not an ideal request for pan capability which does not make much sense.
- Or should PTZ constraints in the advanced constraint sets be ignored or forbidden in `getUserMedia()` altogether? That might actually make most sense as advanced PTZ constraints in `getUserMedia()` do not really have much value. If that route is selected, should PTZ constraints in the advanced constraint sets be silently ignore, or should they produce a console warning or equivalent or should they throw an `OverconstrainedError`?

guidou:
> * What should `getUserMedia({video: {advanced: [{pan: {exact: 1234}, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: 1234, width: 10}]}` do? Should the advanced constraint set be ignored? Note that @youennf's PR [w3c/mediacapture-main#707](https://github.com/w3c/mediacapture-main/pull/707) does not do that. If the advanced constraint set is ignored, should it be ignored silently, should it produce a console warning or equivalent or should it throw an `OverconstrainedError` (which advanced constraints do not normally throw)?

The whole advanced set should be ignored because it contains a required constraint for pan (naked values in advanced are exact). Thus, the examples given should be equivalent to getUserMedia({video:true}).


> * What should `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` do? Is that ideal request for pan capability and for the pan setting of 1234?
Ideal values are ignored for advanced sets. I believe this is equivalent to 
getUserMedia({video: {advanced: [{pan: {}, width: 10}]}), which should be equivalent to getUserMedia({video: {advanced: [{width: 10}]})

> * What should `getUserMedia({video: {advanced: [{pan: true, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: {}, width: 10}]}` do? Is that ideal request for pan capability (like `getUserMedia({video: {pan: true}})`)? Makes kind of sense. Or is that no constraint and no capability request? That would be easy but in that case is  `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` an ideal request for the pan setting of 1234 but not an ideal request for pan capability which does not make much sense.

This should result in no constraints and no (ideal) capability request. There are no ideal values in advanced sets.


> * Or should PTZ constraints in the advanced constraint sets be ignored or forbidden in `getUserMedia()` altogether? That might actually make most sense as advanced PTZ constraints in `getUserMedia()` do not really have much value. If that route is selected, should PTZ constraints in the advanced constraint sets be silently ignore, or should they produce a console warning or equivalent or should they throw an `OverconstrainedError`?

Based on the conclusion in the virtual interim, PTZ constraints in the advanced sets can only have two meanings.
1. They cause the set to be ignored if they include any required constraint (e.g., pan: 4, pan: {min: 4}, pan: {exact: 4}) since we required constraints for PTZ are not allowed.
2. They are ignored because they specify no constraints, so they don't affect anything in the SelectSettings algorithm (e.g., pan: {} or pan: true).

Thus, in practice PTZ constraints in the advanced set are not useful in practice. The reason is that we don't want required constraints for PTZ and the only thing advanced sets can do is specify required constraints.

eehakkin:
I updated the commit and the intro.

> > * What should `getUserMedia({video: {advanced: [{pan: {exact: 1234}, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: 1234, width: 10}]}` do? Should the advanced constraint set be ignored? Note that @youennf's PR [w3c/mediacapture-main#707](https://github.com/w3c/mediacapture-main/pull/707) does not do that. If the advanced constraint set is ignored, should it be ignored silently, should it produce a console warning or equivalent or should it throw an `OverconstrainedError` (which advanced constraints do not normally throw)?
> 
> The whole advanced set should be ignored because it contains a required constraint for pan (naked values in advanced are exact). Thus, the examples given should be equivalent to getUserMedia({video:true}).

Done.

> > * What should `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` do? Is that ideal request for pan capability and for the pan setting of 1234?
> >   Ideal values are ignored for advanced sets. I believe this is equivalent to
> >   getUserMedia({video: {advanced: [{pan: {}, width: 10}]}), which should be equivalent to getUserMedia({video: {advanced: [{width: 10}]})
> > * What should `getUserMedia({video: {advanced: [{pan: true, width: 10}]}` or  `getUserMedia({video: {advanced: [{pan: {}, width: 10}]}` do? Is that ideal request for pan capability (like `getUserMedia({video: {pan: true}})`)? Makes kind of sense. Or is that no constraint and no capability request? That would be easy but in that case is  `getUserMedia({video: {advanced: [{pan: {ideal: 1234}, width: 10}]}` an ideal request for the pan setting of 1234 but not an ideal request for pan capability which does not make much sense.
> 
> This should result in no constraints and no (ideal) capability request. There are no ideal values in advanced sets.

These result in ideal constraints and ideal capability constraints in advanced sets which do not effect the result of the SelectSettings algorithm thus basically the same thing.

> Based on the conclusion in the virtual interim, PTZ constraints in the advanced sets can only have two meanings.
>
> 1. They cause the set to be ignored if they include any required constraint (e.g., pan: 4, pan: {min: 4}, pan: {exact: 4}) since we required constraints for PTZ are not allowed.
> 2. They are ignored because they specify no constraints, so they don't affect anything in the SelectSettings algorithm (e.g., pan: {} or pan: true).
> 
> 
> Thus, in practice PTZ constraints in the advanced set are not useful in practice. The reason is that we don't want required constraints for PTZ and the only thing advanced sets can do is specify required constraints.

Done.

index.bs
jan-ivar: > To make it clear, allowing pan:{} or pan:true to mean a required capability request in an advanced set allows fingerprinting by adding additional constraints and checking if the other constraints were ignored.

@guidou This is after the user has granted permission, so do we care? Can't they already check whether they got ptz permission? If they were, they know the user has a ptz device. If they weren't, then most likely not¹.

> ... which is basically equivalent to having an OverconstrainedError if the capability request was required in the basic set.

They're not equivalent, because `OverconstrainedError` is thrown before any permission prompt, which can be used by tracking libraries without a user's knowledge or permission.

`advanced` constraints cannot throw `OverconstrainedError` by design, so I don't see a significant fingerprint risk.

---
<sub>1. Sure some user agents may offer users a way to downgrade requests to non-ptz, but I doubt many would.</sub>

jan-ivar (changes requested):
I don't see a fingerprinting issue with `advanced` constraints, and I don't see any mention of it in the minutes, so I'm marking this as needs changing. Happy to discuss.

guidou:
Allowing required constraints in advanced sets has the same fingerprinting consequences as allowing them in the basic set.
If they were allowed in the basic set:
getUserMedia({video: {pan: {exact: 10}}) gives OverconstrainedError if pan is not supported, which is a fingerprinting issue.

Similarly, with advanced constraints and assuming 10 is never a default width,
getUserMedia({video: advanced: [{pan: 10, width: 10, resizeMode: "crop-and-scale"}]) gives you width 10 if pan is supported and width different from 10 if pan is not supported, so you get the exact same fingerprinting as in the basic set (note: naked values in advanced are all exact).

jan-ivar:
I also recommend we switching to patching algorithms¹ for greater specificity, since constraints are used in both `getUserMedia` and `applyConstraints`, and the current prose makes it hard to tell where it applies (`getUserMedia` only right?)

As to how to write input validation steps to throw on `exact` ptz constraints, there's precedent in [getDisplayMedia](https://w3c.github.io/mediacapture-screen-share/#dom-mediadevices-getdisplaymedia), so I would look at that:

_"4. For each existing member in constraints whose value, CS, is a dictionary, run the following steps: ... If CS contains a member whose name specifies a constrainable property ..., and whose value in turn is a dictionary containing a member named either min or exact, return a promise rejected with a newly created TypeError."_

We should probably also throw `TypeError`, not `OverconstrainedError`, to be consistent with that precedent.

---
<sub>1. I did some algorithm patching a while ago [here](https://w3c.github.io/webrtc-identity/identity.html#creating-an-offer), if you need an example.</sub>

guidou:
> I also recommend we switching to patching algorithms¹ for greater specificity, since constraints are used in both `getUserMedia` and `applyConstraints`, and the current prose makes it hard to tell where it applies (`getUserMedia` only right?)
> 
> As to how to write input validation steps to throw on `exact` ptz constraints, there's precedent in [getDisplayMedia](https://w3c.github.io/mediacapture-screen-share/#dom-mediadevices-getdisplaymedia), so I would look at that:
> 
> _"4. For each existing member in constraints whose value, CS, is a dictionary, run the following steps: ... If CS contains a member whose name specifies a constrainable property ..., and whose value in turn is a dictionary containing a member named either min or exact, return a promise rejected with a newly created TypeError."_
> 
> We should probably also throw `TypeError`, not `OverconstrainedError`, to be consistent with that precedent.
> 

I agree

jan-ivar:
> Allowing required constraints in advanced sets has the same fingerprinting consequences as allowing them in the basic set.

@guidou Not really, because 

> getUserMedia({video: {pan: {exact: 10}}) gives OverconstrainedError if pan is not supported, which is a fingerprinting issue.

...throws without a prompt (albeit at the risk of prompting the user, but it works even if they deny), whereas

> getUserMedia({video: advanced: [{pan: 10, width: 10, resizeMode: "crop-and-scale"}]) gives you width 10 if pan is supported and width different from 10 if pan is not supported

...only fingerprints **_if_** the user grants permission. If they **_don't_** grant permission then no fingerprint; a significant difference.

guidou:
> > Allowing required constraints in advanced sets has the same fingerprinting consequences as allowing them in the basic set.
> 
> @guidou Not really, because
> 
> > getUserMedia({video: {pan: {exact: 10}}) gives OverconstrainedError if pan is not supported, which is a fingerprinting issue.
> 
> ...throws without a prompt (albeit at the risk of prompting the user, but it works even if they deny), whereas
> 
> > getUserMedia({video: advanced: [{pan: 10, width: 10, resizeMode: "crop-and-scale"}]) gives you width 10 if pan is supported and width different from 10 if pan is not supported
> 
> ...only fingerprints **_if_** the user grants permission. If they **_don't_** grant permission then no fingerprint; a significant difference.

I don't see the difference. Throwing OverconstrsainedErro or returning a track with width 640  (both without prompt) looks like the same fingerprinting to me. Am I missing something?

jan-ivar:
> I don't see the difference. Throwing OverconstrsainedErro or returning a track with width 640  (both without prompt) looks like the same fingerprinting to me. In both cases you detect that the constraint set with the required pan constraint couldn't be satisfied. Am I missing something?

@guidou Most sites on the web do not have camera or microphone permission.

Returning a track requires permission whereas OverconstrainedError does not.

Fingerprinting-wise I'd say that's a big difference. Wouldn't you?

youennf:
> Fingerprinting-wise I'd say that's a big difference. Wouldn't you?

This makes a difference. But don't we want to deprecate advanced constraints?

guidou:
> > I don't see the difference. Throwing OverconstrsainedErro or returning a track with width 640  (both without prompt) looks like the same fingerprinting to me. In both cases you detect that the constraint set with the required pan constraint couldn't be satisfied. Am I missing something?
> 
> @guidou Most sites on the web do not have camera or microphone permission.
> 
> Returning a track requires permission whereas OverconstrainedError does not.
> 
> Fingerprinting-wise I'd say that's a big difference. Wouldn't you?

I see your point now. You get the prompt for the regular camera permission with advanced, which is indeed a significant difference.

jan-ivar:
> This makes a difference. But don't we want to deprecate advanced constraints?

@youennf I like my ducks in a row but not tied together. 😉

eehakkin:
So do I understand correctly that the consensus is to allow required PTZ constraints in the advanced sets, after all?

youennf:
That is what mediacapture-main is mandating with https://github.com/w3c/mediacapture-main/pull/707.
I am not exactly sure mediacapture-image should add normative statements on top of mediacapture-main, but it seems good to describe the intent and behavior.

As of the exact behavior, I would tend to remove advanced constraints ;)

index.bs
```diff
@@ -789,35 +792,67 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+      <dfn>Pan capability</dfn> is a boolean camera property representing camera's ability to pan which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/pan}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw an {{OverconstrainedError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
+        <li>The algorithm MUST ignore the constraint set containing the constraint if the constraint is in the advanced constraints and is <a>required</a>.</li>
```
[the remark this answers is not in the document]
eehakkin: Removed.

index.bs
```diff
@@ -798,35 +801,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+      <dfn>Pan capability</dfn> is a boolean camera property representing camera's ability to pan which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/pan}} dictionary member exists after a possible normalization.
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/pan}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw a {{TypeError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
```
jan-ivar, line 812: I worry that relying on definitions like "required" are unnecessary indirections here.

As I mentioned in https://github.com/w3c/mediacapture-image/pull/261#issuecomment-698514049 I'd much prefer this worded like we do in screen-capture, based solely on inputs.

jan-ivar (changes requested):

index.bs
```diff
@@ -798,35 +801,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+        <li>The algorithm MUST <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId unless there are no pan capable cameras.
```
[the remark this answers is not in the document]
eehakkin: That's the intention, `applyConstraints` never changes exposure or availability of PTZ capabilities thus there is not point for `applyConstraints` to **request permission**.

index.bs
```diff
@@ -798,35 +801,64 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+    <li>
+      <dfn>Zoom</dfn> is a numeric camera setting that controls the focal length of the lens. The setting usually represents a ratio, e.g. 4 is a zoom ratio of 4:1. The minimum value is usually 1, to represent a 1:1 ratio (i.e. no zoom).
 
-    If the {{visibilityState}} of the <a>top-level browsing context</a> value is "hidden", the {{applyConstraints()}} algorithm MUST throw a {{SecurityError}} if {{MediaTrackConstraintSet/zoom}} dictionary member exists after a possible normalization.
+      <dfn>Zoom capability</dfn> is a boolean camera property representing camera's ability to zoom which results in a proper non-degenerate setting range.
+      That capability or whether it is exposed cannot be changed with {{applyConstraints()}}.
 
+      In the {{getUserMedia()}} algorithm, if the {{MediaTrackConstraintSet/zoom}} dictionary member of a {{MediaTrackConstraintSet}} object exists after a normalization and the algorithm uses that dictionary member:
+      <ul>
+        <li>The algorithm MUST throw a {{TypeError}} if the constraint is in the basic constraints and is <a>required</a>.</li>
```
youennf, line 855: Again, this should be left to mediacapture-main (which currently throws OverConstrained error, we could change it there to TypeError).
A note clarifying the intent is fine though.

eehakkin: I assume that you refer to https://github.com/w3c/mediacapture-main/pull/707. I agree that once that is merged, there is no point to specify this here.
Also, https://github.com/w3c/mediacapture-main/pull/707 applies to all Image Capture constraints so a single note in the beginning of the section "10. Photo Capabilities and Constrainable Properties" would probably be sufficient.

index.bs
eehakkin: In https://github.com/w3c/mediacapture-main/pull/707#discussion_r495826772 @youennf seems to suggest the screen-capture to be changed instead. I tend to somewhat agree that copying definitions inline creates unnecessary duplication.

index.bs
```diff
@@ -596,9 +596,10 @@ This Section defines a number of new set of <a>Constrainable Properties</a> for
   <dd>See <a>contrast</a> constrainable property.</dd>
 
   <dt><dfn dict-member for="MediaTrackConstraintSet"><code>pan</code></dfn></dt>
-  <dd>A value of true is normalized to a value of empty {{ConstrainDouble}}.
+  <dd>A value of true is normalized to a value of empty {{ConstrainDoubleRange}}.
       A value of false is normalized to a value of undefined.
-      See <a>pan</a> constrainable property.</dd>
+      If the value after normalization is a double value or a {{ConstrainDoubleRange}} value (whether empty or not), that implies a <a>pan capability</a> request which is ideal in basic constraints and exact <a>required</a> in advanced constraints.
```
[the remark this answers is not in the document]
eehakkin: > Not happy with this sentence, because it doesn't appear to add anything normative that I can see (did you intend it to?) Please remove.

It should normatively state that `navigator.mediaDevices.getUserMedia({video: {pan: true}})` gives lower (better) fitness distance for pan capable cameras than for pan incapable camers and that `navigator.mediaDevices.getUserMedia({video: {height: 480, width: 640, advanced: [{height: 2160, pan: true, tilt: true, width: 3840}]}})` resolves either with a pan and tilt capable UHD video track or with a VGA video track without pan or tilt support.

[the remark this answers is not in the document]
jan-ivar: @eehakkin I don't think it accomplishes that, because a _"implies a pan capability request"_ is not a thing in the constraints algorithm. I think we need https://github.com/w3c/mediacapture-image/pull/257 to fix this.

index.bs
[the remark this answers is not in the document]
jan-ivar: > That's the intention, `applyConstraints` never changes exposure or availability of PTZ capabilities thus there is no point for `applyConstraints` to **request permission**.

@eehakkin I see no support in the minutes for this normative change. Did I miss it?

From the current spec, a site may access a camera through regular permission, discover it has `pan: true` capabilites thanks to https://github.com/w3c/mediacapture-image/pull/260, and then use `track.applyConstraints({pan: true})` to request more (ptz) permission.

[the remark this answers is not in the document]
youennf: > Did I miss it?

This was discussed so the recording should have it.
The idea was: regular getUserMedia, check capabilities, and then PTZ getUserMedia to get PTZ control.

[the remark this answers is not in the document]
jan-ivar: @youennf Is the recording available yet? I cannot seem to find it. In any case, consensus at w3c is written not verbal, and I see no resolution taken.

> The idea was: regular getUserMedia, check capabilities, and then PTZ getUserMedia to get PTZ control.

Then why disallow: regular getUserMedia, check capabilities, and then PTZ applyConstraints to get PTZ control?

Sites calling `getUserMedia` again when they could be using `applyConstraints` is an anti-pattern. It encourages sites to structure their code in a way that frequently causes extra prompts all over the web for Firefox users who grant permission as one-shots by default (because there's a competing incentive to stop tracks before gUM to avoid limitations on phones).

While I understand that here we think this is not a problem because ptz will always prompt again, I think that's a faulty assumption. It's conceivable a user agent might allow a user to opt into always granting ptz with regular camera requests, but still gate camera requests behind one-time permissions (i.e. _"always ask me for camera, but ptz is fine"_).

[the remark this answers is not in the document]
youennf: > Then why disallow: regular getUserMedia, check capabilities, and then PTZ applyConstraints to get PTZ control?

I previously backed up the idea of this flow and I am happy to discuss this further but I feel this is a separate issue.
This PR discussion is very long and it feels like, to make progress, we should split the PR and solely focus here on the initial scope: "Reject with TypeError on required PTZ contraints in basic getUserMedia contraints"

index.bs
jan-ivar: @eehakkin https://github.com/w3c/mediacapture-main/pull/707 has merged, so lets remove this!

index.bs
jan-ivar: @eehakkin https://github.com/w3c/mediacapture-main/pull/707 has been merged now, so lets remove this sentence.

jan-ivar:
@eehakkin is this PR necessary anymore or can it be closed?

index.bs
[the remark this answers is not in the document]
jan-ivar: @youennf with https://github.com/w3c/mediacapture-main/pull/707 and https://github.com/w3c/mediacapture-image/pull/264 merged is there anything left here we need? Can we close it?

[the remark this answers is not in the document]
youennf: We could have non normative text that tells that required constraints will throw.
We have #264 but this is not in the spec.
I do not feel strongly either way.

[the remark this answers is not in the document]
eehakkin: I think that it warrants a note but a general one and not a PTZ specific one. I'll create a separate issue for that.

index.bs
eehakkin: Yes, it is not needed anymore.

eehakkin:
> @eehakkin is this PR necessary anymore or can it be closed?

The type error issue is now handled. Let's close this one and let's handle other issues separately.

PR #265, opened 2020-09-30
Fix nits in PTZ example

beaufortfrancois:
This PR fixes nits in PTZ example:
- Use `settings` instead of `capabilities` for checking whether PTZ is available.
- Use `disabled` instead of custom CSS class `hidden` for clarity
- Remove use of template literal as bikeshed doesn't like it. See current state at https://w3c.github.io/mediacapture-image/#example1

image

@riju Please review and merge.


riju (approved):
:+1:

Issue #279, opened 2021-05-04
Should check results of PTZ permission requests


eehakkin:
The spec requests `{name: "camera", panTiltZoom: true}` permissions but does not check whether the permissions were granted or denied. That is clearly not the intention.

PR #280, opened 2021-05-04
Check the results of PTZ permission requests

eehakkin:
As stated in #279, the spec requests `{name: "camera", panTiltZoom: true}` permissions but does not check whether the permissions were granted or denied. That is clearly not the intention.


riju (approved):
Let's wait for @jan-ivar 's blessing.

index.bs
```diff
@@ -805,15 +805,15 @@ When the {{getSettings()}} method is invoked on a video stream track, the user a
...
+    Any algorithm which uses a {{MediaTrackConstraintSet}} object whose {{MediaTrackConstraintSet/pan}} dictionary member exists with a value other than false MUST either <a>request permission to use</a> (as defined in [[!permissions]]) a PermissionDescriptor with its name member set to {{PermissionName/camera}} and its {{CameraDevicePermissionDescriptor/panTiltZoom}} member set to true, and, optionally, consider its {{DevicePermissionDescriptor/deviceId}} member set to any appropriate device's deviceId, and check that the permission is granted or decide not to expose the pan setting in <a>settings dictionaries</a> or in {{MediaTrackSettings}} dictionaries.
```
jan-ivar, line 808: This seems to say that the only penalty for not having permission is that `pan` will not be exposed in `getSettings`, which we already have superior language for [here](https://w3c.github.io/mediacapture-image/#dom-mediatracksettings-pan).

The sentence is also running long, making the and/or logic hard to follow. 

Instead, how about:  "If permission is not granted, the algorithm MUST ignore the member".

jan-ivar (changes requested):

index.bs
jan-ivar: ...in order to burn out the ability to constrain on it.

jan-ivar (approved):

